PR #118, opened 2013-12-08
Add HV furnace and grinder (by cheapie)

qwrwed:
https://forum.minetest.net/viewtopic.php?id=7953


cheapie:
This has been replaced by #201.

PR #131, opened 2014-02-16
Add button to toggle Technic machine protection on and off

Bremaweb:


asl97:
reposting from issue 191 (i didn't link it cause it's already link above)

a13e7b7 doesn't include support for LV

PR #157, opened 2014-09-05
don't use dummy textures for the template replacer and tool

HybridDog:
When I searched stuff in ui, the error message annoyed me.


HybridDog:
l put the commit there:
https://github.com/minetest-technic/technic/pull/249

PR #167, opened 2014-12-01
Add gloopores' Kalite dust as a more effecient fuel than lumps

thatgraemeguy:
This PR replaces #162 which I messed up nicely :-)


thatgraemeguy:
Kalite dust only exists because its defined in https://github.com/minetest-technic/technic/blob/master/technic/machines/register/grinder_recipes.lua

ShadowNinja:
Ah, seems O.K. then.

ShadowNinja:
2e04374de8bd0e39850e0454742e3c411fab6dac

PR #174, opened 2015-01-10
support more nodes to be chainsawn

t4im:
completes tree/leave/fruit nodes of default, moretrees, farming_plus
and adds support for vines (generated by moretrees if available) and snow


ShadowNinja:
0114feaf04e54ca711b9ce7770b322e1550472a7

PR #175, opened 2015-01-10
reuse the new screwdriver code of minetest/minetes_game in adapted form ...

t4im:
...for the sonic screwdriver


Novatux:
Could it be possible to call the code from _game instead of having to rewrite it, so that changes to original code would change that code too?

t4im:
Not without changing minetest_game code first.
It uses `local` functions uses an additional `can_dig` check that we might want to skip (to continue to allow rotating CLU contained machines for example) and of course has a different means of wearing.

VanessaE:
fe4372649fd7a265cbacf18f0d29032d958aa24c

PR #177, opened 2015-01-14
use an alternative way to "overwrite" the pipeworks tube-recipes

t4im:
This is an attempt of a different form of overwriting the recipes of pipeworks.
The former way was -- although inactive -- listed still in the crafting menu, leading people to try to craft these via autocrafters. Autocrafters do correctly craft these technic:nothing nodes, which are effective inventory-DoS items, as they cannot be differentiated from empty slots (unless stacked).

This is an alternative approach, that allows people to craft via these recipes but only returns the material as used back to them in both inventory as well as autocrafter.

I tried to get an opinion on this in irc but couldn't get any; thus this proactive question of this solution in form of a pullrequest.

### DIFF
--- a/technic/crafts.lua
+++ b/technic/crafts.lua
@@ -159,30 +159,16 @@ minetest.register_craft({
 })
 
 -- Remove some recipes
-minetest.register_craftitem("technic:nothing", {
-	description = "",
-	inventory_image = "blank.png",
-	groups = { not_in_creative_inventory = 1 },
-})
-
-if minetest.register_craft_predict then
-	minetest.register_craft_predict(function(itemstack, player, old_craft_grid, craft_inv)
-		if itemstack:get_name() == "technic:nothing" then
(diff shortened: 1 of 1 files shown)


Novatux:
This has the drawback of destroying items if some input stacks contain more than one item. Actually, the real problem lies in autocrafters, which should be made to work as other pipeworks machines, creating a fake player and actually calling the functions.

t4im:
Oh… indeed. Isn't this also a minetest bug then? Can't imagine item destruction being intentional on replacement recipes.

t4im:
While working on recipes in another place, i came up with yet another alternative way… this time a lot cleaner. I'm mildly embarassed not to have tried this the first time. ;)

Novatux:
https://github.com/minetest-technic/technic/commit/7e92fe0a46bdc97c39129918ddc226d03862f68d

PR #178, opened 2015-01-16
Fix support for acacia trees

Wayward1:
Just realized I had the wrong item string >.<


VanessaE:
9e1c5740d3fa79ea8a8dffc658c2db6d798a4a85

PR #179, opened 2015-01-19
diverse recipe additions and enhancements

t4im:


thatgraemeguy:
The extraction of dye from flowers still suffers from the same problem it always has, i.e. using an extractor yields exactly the same amount as plain crafting. Granted, it used to be worse as an extractor would yield only half the amount of dye that plain crafting does, but this change is still not giving the player any incentive to use an extractor rather than just plain crafting.

Perhaps technic can override the default recipe for crafting dye so that those only yield 2 dye per flower, thus incentivising the player to use the extractor to increase yield.

t4im:
It does actually almost exactly that ;-) starting from https://github.com/t4im/technic/blob/1d1b51eaed34cf0006184b18266f115547a24faa/technic/machines/register/extractor_recipes.lua#L49

thatgraemeguy:
Oh, well..... yes..... mmmm.... ok then.

I shall endeavor to read more completely in future. :blush: 

:+1:

t4im:
rebased after a few more additions and changes

technic/machines/register/grindings.lua
```diff
@@ -0,0 +1,64 @@
...
+local moretrees = minetest.get_modpath("moretrees")
+local mesecons_materials = minetest.get_modpath("mesecons_materials")
+local dye = minetest.get_modpath("dye")
+
+-- sawdust, the finest wood/tree grinding
+local sawdust = "technic:sawdust"
+minetest.register_craftitem(sawdust, {
+	description = S("Sawdust"),
+	inventory_image = "technic_sawdust.png",
+	on_place_on_ground = minetest.craftitem_place_item,
```
MT-Modder: I don't believe ```on_place_on_ground``` is a valid callback. ```on_place``` is though. I could be wrong.

technic/machines/register/grindings.lua
```diff
@@ -0,0 +1,64 @@
...
+	lname = string.gsub(lname, ' ', '_')
+	local grindings_name = "technic:"..lname.."_grindings"
+	local inventory_image = "technic_"..lname.."_grindings.png"
+	if grinding_color then
+		inventory_image = inventory_image .. "^[colorize:" .. grinding_color
+	end
+	minetest.register_craftitem(grindings_name, {
+		description = S("%s Grinding"):format(S(name)),
+		inventory_image = inventory_image,
+		on_place_on_ground = minetest.craftitem_place_item,
```
MT-Modder: Same here.

t4im:
yea that's some harmless copy pasta from the ore-dust code i did yesterday in tiredness to avoid searching the engine for possible undocumented functions (they sadly exist)… 
did so this morning, and you're right, its not being called back anywhere (or at least not anymore)
I'll remove it with another pullrequest containing smaller improvements later.

MT-Modder:
Okay, I wasn't sure because I know there are some undocumented methods that I would never know about :)

PR #180, opened 2015-01-24
Quarry optimisations/improvements

est31:
Makes quarries less resource-intensive. Don't forceload the whole digging range every time, also add a quarry cache that collects items and outputs them in stacks instead of single items.

### DIFF
--- a/technic/helpers.lua
+++ b/technic/helpers.lua
@@ -36,3 +36,11 @@ function technic.function_exists(function_name)
 	return type(resolve_name(function_name)) == 'function'
 end
 
+-- if the node is loaded, returns it. If it isn't loaded, load it and return nil.
+function technic.get_or_load_node(pos)
+	local node_or_nil = minetest.get_node_or_nil(pos)
+	if node_or_nil then return node_or_nil end
+	local vm = VoxelManip()
+	local MinEdge, MaxEdge = vm:read_from_map(pos, pos)
+	return nil
+end
--- a/technic/machines/HV/quarry.lua
+++ b/technic/machines/HV/quarry.lua
@@ -108,21 +108,26 @@ local function quarry_run(pos, node)
 				vector.multiply(pdir, rp)),
 				vector.multiply(qdir, rq))
 			local can_dig = true
-			for ay = startpos.y, digpos.y+1, -1 do
-				if data[area:index(digpos.x, ay, digpos.z)] ~= c_air then
-					can_dig = false
-					break
-				end
-			end
 			if can_dig and minetest.is_protected and minetest.is_protected(digpos, owner) then
 				can_dig = false
 			end
 			local dignode
 			if can_dig then
-				dignode = minetest.get_node(digpos)
+				dignode = technic.get_or_load_node(digpos) or minetest.get_node(digpos)
 				local dignodedef = minetest.registered_nodes[dignode.name] or {diggable=false}
 				if not dignodedef.diggable or (dignodedef.can_dig and not dignodedef.can_dig(digpos, nil)) then
 					can_dig = false
+				end
+			end
+
+			if can_dig then
+				for ay = startpos.y, digpos.y+1, -1 do
+					local checkpos = {x=digpos.x, y=ay, z=digpos.z}
+					local checknode = technic.get_or_load_node(checkpos) or minetest.get_node(checkpos)
+					if checknode.name ~= "air" then
+						can_dig = false
+						break
+					end
 				end
 			end
 			nd = nd + 1
--- a/technic/machines/HV/quarry.lua
+++ b/technic/machines/HV/quarry.lua
@@ -209,6 +209,9 @@ minetest.register_node("technic:quarry", {
 	tiles = {"technic_carbon_steel_block.png", "technic_carbon_steel_block.png",
 	         "technic_carbon_steel_block.png", "technic_carbon_steel_block.png",
 	         "technic_carbon_steel_block.png^default_tool_mesepick.png", "technic_carbon_steel_block.png"},
+	inventory_image = minetest.inventorycube("technic_carbon_steel_block.png",
+	         "technic_carbon_steel_block.png^default_tool_mesepick.png",
+	         "technic_carbon_steel_block.png"),
 	paramtype2 = "facedir",
 	groups = {cracky=2, tubedevice=1, technic_machine = 1},
 	tube = {
--- a/technic/machines/HV/quarry.lua
+++ b/technic/machines/HV/quarry.lua
@@ -15,23 +15,25 @@ local quarry_demand = 10000
 
 local function set_quarry_formspec(meta)
 	local radius = meta:get_int("size")
-	local formspec = "size[6,2.5]"..
-		"item_image[0,0;1,1;technic:quarry]"..
-		"label[1,0;"..S("%s Quarry"):format("HV").."]"..
-		"field[0.3,1.5;2,1;size;"..S("Radius:")..";"..radius.."]"
+	local formspec = "size[6,4.3]"..
+		"list[context;cache;0,1;4,3;]"..
+		"item_image[4.8,0;1,1;technic:quarry]"..
+		"label[0,0.2;"..S("%s Quarry"):format("HV").."]"..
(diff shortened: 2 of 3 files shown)


est31:
Added commit that gives a nicer inventory icon.

technic/helpers.lua
```diff
@@ -36,3 +36,10 @@ function technic.function_exists(function_name)
 	return type(resolve_name(function_name)) == 'function'
 end
 
+function technic.load_position(pos)
+	local node_or_nil = minetest.get_node_or_nil(pos)
+	if node_or_nil then return node_or_nil end
+	local vm = VoxelManip()
+	local MinEdge, MaxEdge = vm:read_from_map(pos, pos)
+	return nil
+end
```
ShadowNinja: 1. I'm pretty sure this function already exists.
  2. `get_node_or_nil` has existed for longer than technic cares to support.  I've never heard of it loading anything though.

technic/machines/HV/quarry.lua
```diff
@@ -64,14 +66,50 @@ local function quarry_receive_fields(pos, formname, fields, sender)
...
+		if stack then
+		local item = stack:to_table()
+		if item then
+			technic.tube_inject_item(pos, pos, vector.new(0, 1, 0), item)
+			stack:clear()
+			inv:set_stack("cache", i, stack)
+			break
+			end
+		end
+	end
```
ShadowNinja: Indentation is messed up.

technic/machines/HV/quarry.lua
```diff
@@ -165,7 +224,18 @@ minetest.register_node("technic:quarry", {
...
+									return 0
+									end,
+	allow_metadata_inventory_put = function(pos, listname, index, stack, player)
+									minetest.chat_send_player(player:get_player_name(), S("Manually taking/removing from cache by hand is not possible. If you can't wait, restart or disable the quarry to start automatic purge."))
+									return 0
+									end,
+	allow_metadata_inventory_take = function(pos, listname, index, stack, player)
+									minetest.chat_send_player(player:get_player_name(), S("Manually taking/removing from cache by hand is not possible. If you can't wait, restart or disable the quarry to start automatic purge."))
+									return 0
+									end
```
ShadowNinja: 1. Duplication.
  2. Waaaaaay too much indentation.

technic/machines/switching_station.lua
```diff
@@ -82,15 +82,9 @@ local add_new_cable_node = function(nodes, pos)
 	return true
 end
 
-local load_position = function(pos)
-	if minetest.get_node_or_nil(pos) then return end
```
ShadowNinja: Ah, I see, you modified this.  Your modified version is incorrect.  This uses `get_node_or_nil` to check if the position is loaded already, yours uses it as the loader.

technic/machines/HV/quarry.lua
```diff
@@ -64,14 +66,50 @@ local function quarry_receive_fields(pos, formname, fields, sender)
...
+
 local function quarry_run(pos, node)
 	local meta = minetest.get_meta(pos)
-	if meta:get_int("enabled") and meta:get_int("HV_EU_input") >= quarry_demand then
+	local inv = meta:get_inventory()
+	-- set cache and purge values if we load an older world
+	inv:set_size("cache", 12)
+	if not meta:get_int("purge_on") then
+		meta:set_int("purge_on", 0)
+	end
```
ShadowNinja: The default for ints is 0, so this isn't needed (default for strings is the empty string).

technic/machines/HV/quarry.lua
```diff
@@ -64,14 +66,50 @@ local function quarry_receive_fields(pos, formname, fields, sender)
...
+end
+
 local function quarry_run(pos, node)
 	local meta = minetest.get_meta(pos)
-	if meta:get_int("enabled") and meta:get_int("HV_EU_input") >= quarry_demand then
+	local inv = meta:get_inventory()
+	-- set cache and purge values if we load an older world
+	inv:set_size("cache", 12)
```
ShadowNinja: This is limited and might overflow.

technic/machines/HV/quarry.lua
```diff
@@ -64,14 +66,50 @@ local function quarry_receive_fields(pos, formname, fields, sender)
...
+	if not meta:get_int("purge_on") then
+		meta:set_int("purge_on", 0)
+	end
+	-- toss a coin whether we do an automatic purge. Chance 1:200
+	local purge_rand = math.random()
+	if purge_rand <= 0.005 then
```
ShadowNinja: I don't like the idea of a random purge.  It could take a second or an hour.

est31: random purges are to prevent quarries that got placed close to each other to not purge simultaneously, and cause cpu and material spikes.

technic/machines/HV/quarry.lua
```diff
@@ -108,33 +146,51 @@ local function quarry_run(pos, node)
...
+					end
+				end
+			end
 			nd = nd + 1
 			if can_dig then
 				minetest.remove_node(digpos)
-				for _, item in ipairs(minetest.get_node_drops(dignode.name, "")) do
-					technic.tube_inject_item(pos, pos, vector.new(0, 1, 0), item)
+				local drops = minetest.get_node_drops(dignode.name, "")
+				local oflow = 99 --just to be sure...
```
est31: TODO: remove

technic/machines/HV/quarry.lua
est31: Thats what [this if clause](https://github.com/minetest-technic/technic/pull/180/files#diff-5e831e33fc7ea0821eb5d89bd2e935b4R178) is for. Although I see that it doesn't 100% fulfill its purpose.

technic/machines/switching_station.lua
est31: It is correct AFAICT. if the node is loaded, I dont need to load, and return get_node_or_nil in order to not call it twice. However, I wanted to spare a second call in that method, as its used in switching station, and there the return value doesnt matter, so I return nil, and fetch the block inline.

est31:
amended following @ShadowNinja 's suggestions.

est31:
Made quarries with empty inventories undiggable

technic/machines/HV/quarry.lua
```diff
@@ -108,43 +143,75 @@ local function quarry_run(pos, node)
...
+		quarry_handle_purge(pos)
 	end
 	set_quarry_formspec(meta)
 	set_quarry_demand(meta)
 end
 
+local function send_move_error(player)
+	minetest.chat_send_player(player:get_player_name(),
+		S("Manually taking/removing from cache by hand is not possible. "..
+		"If you can't wait, restart or disable the quarry to start automatic purge."))
```
ShadowNinja: Why isn't taking from the cache allowed?

technic/machines/switching_station.lua
ShadowNinja: I see what you're doing now.

technic/machines/HV/quarry.lua
```diff
@@ -64,14 +66,47 @@ local function quarry_receive_fields(pos, formname, fields, sender)
...
+				technic.tube_inject_item(pos, pos, vector.new(0, 1, 0), item)
+				stack:clear()
+				inv:set_stack("cache", i, stack)
+				break
+			end
+		end
+	end
+	if inv:is_empty("cache") then
+		meta:set_int("purge_on", 0)
+	end
```
ShadowNinja: You could speed this up by doing `inv:set_list("cache", {})` and not checking if it's empty at the end (it has to be empty).

technic/machines/HV/quarry.lua
est31: The main purpose of the cache is to reduce load on the server to disable. I don't want to make quarries easier to deploy, not requiring at least a chest attached to the quarry. Also, the mining procedure is dangerous, and people shoudn't hurt themselves when they attempt to take stuff from the quarry. :) And the nice formspec would be messed up when I also displayed the player's inventory.
If you think this is not ok, I can make stuff removable from the quarry.

VanessaE:
Seems to behave well enough.

PR #184, opened 2015-01-31
Better number formating

est31:
Is it 120000 or 1200000?


est31:
Using spaces is SI style. International.

PR #185, opened 2015-01-31
Better number formating

est31:
Is it 120000 or 1200000?


VanessaE:
39b2f26ff92374496f2fe466eee4d32397cd2d95

PR #189, opened 2015-02-03
Make technic chests independent from unified inventory

est31:
Copy the textures made by RealBadAngel (under WTFPL) from unified inventory to technic to avoid minetest not finding them when unified inventory is not installed.


est31:
ready for merge :)

PR #191, opened 2015-02-06
machine upgrades and chainsaw fixes

t4im:
* adds an upgrade item to make a machine public (unlocked chest)
* drop upgrade items if dug
* prevent setting unknown upgrades or more than one item into the uprade slots


asl97:
i thought github would at post my comment at #131 instead of here since i put #131

PR #198, opened 2015-02-12
Get rid of technic.format

est31:
SN didnt like it, and it wasn't beautiful (and performant) either. Replaces technic.format, and uses the newest initlib features. Translations can update.


ShadowNinja:
:+1:
I think you should update the translations yourself though, since you don't need a knowledge of the language for a change this simple.

est31:
Changed.

PR #199, opened 2015-02-14
test if stairsplus is available

HybridDog:
approximately the same
https://github.com/Splizard/minetest-mod-snow/pull/8

### DIFF
--- a/extranodes/init.lua
+++ b/extranodes/init.lua
@@ -1,3 +1,7 @@
+if not rawget(_G, "stairsplus") then
+	error("[technic/extranodes] stairsplus not found")
+end
+
 -- Minetest 0.4.6 mod: extranodes
 -- namespace: technic
 -- Boilerplate to support localized strings if intllib mod is installed.


t4im:
:+1: 
also great example for why minetest/minetest#1862 could be useful too ;-)

ShadowNinja:
If `extranodes` depends on `stairsplus` it shouldn't just silently fail if it isn't available.  It should either use a hard dependency or it should throw an error.

HybridDog:
>It should either use a hard dependency

how? stairsplus isn't a mod

> it should throw an error

ok

ShadowNinja:
@HybridDog It's a part of moreblocks.  The error would be a `error()`, meaning it would cause the startup to fail and it would print the error.  That way you can see that there's something wrong and fix it.  Loging to infostream is unacceptable, that's only ever enabled when debuging because so many things are logged there.

HybridDog:
>It's a part of moreblocks.

sorry, l thought moreblocks would offer a setting to disable stairsplus…

PR #201, opened 2015-02-22
Add HV furnace, grinder (from cheapie), and change the behavior of control logic units to be based on machine production rate

BluebirdGreycoat:
This is based on the older pull request here: https://github.com/minetest-technic/technic/pull/118
Note that this _completely_ changes the item-ejection behavior of machines. The reason for the change is that during an IRC conversation, it came up as a prerequisite to HV machines in general, since HV machines would (to avoid jamming) need to eject items at a much faster rate.


technic/machines/register/common.lua
```diff
@@ -74,9 +74,14 @@ function technic.send_items(pos, x_velocity, z_velocity, output_name)
 		if stack then
 			local item0 = stack:to_table()
 			if item0 then 
-				item0["count"] = "1"
+				-- Calculate how many items to move. Don't move more items than exist.
+				-- Basically, 'item_count' is the maximum number of items to move.
+				local num_items = item_count or 1
+				if stack:get_count() < num_items then num_items = stack:get_count() end
+
```
est31: This line ends with spaces. please remove them, its bad practice. Also, its not the only line with that problem.

technic/machines/register/common.lua
```diff
@@ -84,8 +89,11 @@ function technic.send_items(pos, x_velocity, z_velocity, output_name)
 	end
 end
 
-
+-- This function is not used anywhere in Technic. Maybe it *might* be used by
+-- other mods?
 function technic.smelt_item(meta, result, speed)
+	minetest.log("warning", "technic.smelt_item() is not used in Technic.")
```
est31: If we want to deprecate this function, then you should make the warning print "technic.smelt_item() is deprecated."
If we want to find out whether this function is used by a mod, then we can put some `if technic.config:get_bool("enable_smelt_item")` around the function definition, with some comment to notify us that they use the method -- if the change causes an issue, then we hopefully get notified, and the people relying on the function can fix it immediately

est31:
:+1:

thatgraemeguy:
Would be nice to have HV counterparts for all the MV machines. I for one wouldn't bother with just 2 HV machines, I'd rather work toward upgrading and eliminating MV. It's annoying enough already to have to keep an LV cable around for the CNC machine.

ShadowNinja:
The slots should handle stacks of more than one upgrade properly first.  The size should also be increased to one `2*2` inventory or so.

t4im:
upgrades were already adapted to only take a single item per slot, and only known upgradeitems

technic/machines/register/common.lua
```diff
@@ -117,7 +192,7 @@ function technic.handle_machine_pipeworks(pos, tube_upgrade, send_function)
 	local x_velocity = 0
 	local z_velocity = 0
 
-	-- Output is on the left side of the furnace
+	-- Output is on the left side of the machine.
```
t4im: you can remove or replace that comment completely when already touching it, orientation is a bit more complicated than that comment and depends on cardinal direction (in fact in 2 cases it ejects to the right)

technic/machines/register/common.lua
```diff
@@ -128,23 +203,55 @@ function technic.handle_machine_pipeworks(pos, tube_upgrade, send_function)
...
+
+	if tube_time >= 10 then
 		tube_time = 0
+
 		if output_tube_connected then
-			send_function(pos, x_velocity, z_velocity)
+			-- Calculate how many items to eject during this run. Eject none if no
+			-- tube upgrades exist, and eject as many as were produced if the max
+			-- number of tube upgrades exist.
+			local ejection_count = meta:get_int("eject_count")
```
t4im: You already got the value in line 208; no need to get it twice from the database in often called code like this.

technic/machines/register/common.lua
```diff
@@ -173,42 +279,69 @@ function technic.machine_after_dig_node(pos, oldnode, oldmetadata, player)
...
-	local to_upgrade = to_list == "upgrade1" or to_list == "upgrade2"
-	local from_upgrade = from_list == "upgrade1" or from_list == "upgrade2"
 
-	if (not public or to_upgrade or from_upgrade) and minetest.is_protected(pos, playername) then
-		minetest.chat_send_player(playername, S("Inventory move disallowed due to protection"))
+
+-- Determine the inventory protection status of a machine. Returns 'true' if
+-- the machine is protected, 'false' if it is not protected or public.
+local function check_machine_protection(player_name, pos, meta)
+	if minetest.is_protected(pos, player_name) then
```
t4im: In the previous code the protection was intentionally the last thing to check, as you never know how expensive the code might be. But it will always be more expensive than a simple metadata lookup, that you will have to do anyway if it is protected.

VanessaE:
What is the current status of this pull request?

BluebirdGreycoat:
This reply is late due to my being on vacation; sorry.

As far as I'm aware, this pull request is simply waiting for review/approval. But whether it gets accepted or not, it'd be great if a decision could be reached so that I'm not spending weeks in suspense. :)

dancetack2:
i think there should also be hv alloy furnaces and with the logic upgrade, you can control how much of what goes into it, that way you can have automated alloy furnaces without having to worry about one material taking up both slots

thomasrudin:
I'm sorry to revive this thread, but why was this closed without an explanation?

numberZero:
It is definitely oversized.

PR #203, opened 2015-03-01
make tools not wear in creative mode

MT-Modder:


VanessaE:
Rather than doing a minetest.setting_getbool() for each place that you want to check for creative mode, do it once somewhere in the mod's init stage, do whatever sanity checks are needed there as well, cache the result into a global variable (well, put it in the technic. table), and reference that variable where it's needed instead.

MT-Modder:
Okay makes sense to me. I wish I thought about that :smiley:

MT-Modder:
Closing for #204

PR #204, opened 2015-03-01
Tools don't wear in creative mode v2

MT-Modder:
- Removed cans (will add again per request)
- Added creative check to global namespace (checking only once as
requested by VanessaE)


VanessaE:
b8c902868115a0a607da35e208a422e430d46317

PR #212, opened 2015-03-11
Wrench fixes

Rogier-5:
Hi
I combined three fixes to the wrench code in this pull request. Please let me know if you'd prefer three separate pull requests.

Fixes:
* Make wrench usable independently of technic
* When restoring the contents of a container, the rotation of the container is now preserved (it used to be reset to the default rotation)
* When restoring a furnace (and maybe other containters as well?), the contents were not restored correctly. Now they are

FYI: while testing, I noticed that some of the technic containers are can not be picked up. E.g. technic:lv_compressor or technic:lv_grinder. Apparently they used to be called technic:compressor etc, but the wrench doesn't know about the new name, and refuses to pick them up...
Edit: submitted this as issue https://github.com/minetest-technic/technic/issues/213

### DIFF
--- a/wrench/init.lua
+++ b/wrench/init.lua
@@ -36,10 +36,11 @@ end
 
 local function restore(pos, placer, itemstack)
 	local name = itemstack:get_name()
+	local node = minetest.get_node(pos)
 	local meta = minetest.get_meta(pos)
 	local inv = meta:get_inventory()
 	local data = minetest.deserialize(itemstack:get_metadata())
-	minetest.set_node(pos, {name = data.name})
+	minetest.set_node(pos, {name = data.name, param2 = node.param2})
 	local lists = data.lists
 	for listname, list in pairs(lists) do
 		inv:set_list(listname, list)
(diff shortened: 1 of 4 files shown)


VanessaE:
4874e290256c195d2abebf406797aaa2bf42582a  78bfa4f893ef577332b11ba199b980b836ad0879 ab6f2b2556f42e965494e6db82eeb969a7ffe924

wrench/init.lua
```diff
@@ -36,14 +36,12 @@ end
 
 local function restore(pos, placer, itemstack)
 	local name = itemstack:get_name()
+	local node = minetest.get_node(pos)
 	local meta = minetest.get_meta(pos)
 	local inv = meta:get_inventory()
 	local data = minetest.deserialize(itemstack:get_metadata())
-	minetest.set_node(pos, {name = data.name})
+	minetest.set_node(pos, {name = data.name, param2 = node.param2})
```
ShadowNinja: Er, this sets `param2` to the `param2` of the node that was there before (probably `air`) rather than the `param2` that the node had when you picked it up...

Rogier-5: restore() is attached to the after_place_node callback...

ShadowNinja: Oh, I see what you're doing.  You're not restoring the original rotation, you're keeping the new rotation from, eg, `rotate_and_place`.  Nevermind then.

Issue #216, opened 2015-03-16
Server crash while trying to move framed structure


neko259:
```
technic/technic/machines/other/frames.lua:446: attempt to index field 'frame_object' (a nil value)
```

I don't know how to reproduce it.  The only thing I know is that it might be related to water (the entire structure was flooded).  Reproducible every time I try to move this structure.

HybridDog:
l assume the frames became objects and then water flew into them.

VanessaE:
Tried to reproduce this crash with current git.  Created a small structure with a few different items on it, then flooded it with water.  Made sure the space under the frames and "upstream" from the motor was clear of water. Still a problem, though the cause of the crash appears rather different now:

```
01:13:07: WARNING: Assignment to undeclared global "c" inside a function at minetest/mods/technic/technic/machines/other/frames.lua:527.
01:13:07: WARNING: Assignment to undeclared global "l2" inside a function at minetest/mods/technic/technic/machines/other/frames.lua:227.
01:13:07: WARNING: Assignment to undeclared global "l1" inside a function at minetest/mods/technic/technic/machines/other/frames.lua:228.
Segmentation fault (core dumped)
```

HybridDog:
ls it fixed yet?

VanessaE:
A mod shouldn't be able to cause anything more tha a routine crash of the Lua code, let alone a segfault.

HybridDog:
There are many ways to cause segfaults, assertion failures, etc..
You could save a voxelmanip object (the reference) and use it later, e.g. with minetest.after, that causes crashes now and then, and you could use animated tiles textures incorrectly.
l'm wondering if mods can break mod security (heap overflow attacks).

Issue #170, opened 2014-12-14
frames cheat


HybridDog:
https://imgrush.com/TEV8R6pJ-dbY
If you place a frame, then place something into it, you have a node with object which you can use to get infinite nodes. You just need to right click on the frame entity, and not on the node inside the frame.

VanessaE:
Can't reproduce this in current Technic.

t4im:
Never could reproduce this particular one either, but i had a similar duplication issue with templates instead of frames. They certainly need *some* repair. See the list in #217

HybridDog:
l just tested it, it still works, see the video.

HybridDog:
Is this fixed yet?

BobFred7:
The link now redirects to what appear to be scam websites

HybridDog:
> The link now redirects to what appear to be scam websites

Thanks for the information. I've removed it now and tested it again. The problem is still not fixed.

PR #217, opened 2015-03-16
allow activation/deactivation of frames/templates by config

t4im:
have them deactivated by default due to a larger collection of associated issues
including but not limited to
* #170
* #216
* item duplication involving templates
* runaway machines
* traversal loop issues with huge machines leading to lag-outs


VanessaE:
137695e

PR #218, opened 2015-03-17
Chests fixes

Rogier-5:
These patches make technic_chests usable independently of moreores, pipeworks and of technic.


t4im:
There is now the issue, that moreore removal will unregister the moreore based chests, thus render already existing chests in a world unusable (not just their crafting).

Rogier-5:
I see...

You mean that the fact that silver chests can't be crafted due to undefined ingredients, does not necessarily imply that they don't exist in some world ?

That's an oversight...
I'll revise the patch and make sure the silver and mithril chests remain defined.

Rogier-5:
Fixed...

VanessaE:
fcef0ef987f7bd7d26d1c8a4d27a562822b580bc ae7f49de385638c0c0ecc1be791bb044b369f5c7 84812d23078c02538c5b6b7bc837adebd5d5d389

PR #221, opened 2015-03-22
Small networking refactor and leaves bugfix

est31:
see commits.


est31:
added a solar array bugfix.

PR #223, opened 2015-04-11
Add Admin Powernode

Ryan-Nolan:
Adds a non-craftable node that allows a player to mimic the hv power
that you get from a nuclear reactor. its designed for admin use only, In
the future i will make the power level configurable via formspec.


ShadowNinja:
This is clearly an HV machine, so it should be in `machines/HV`.  `machines/other` is for machines that don't use Technic's electrical system.
Also, it seems like if this is useful for HV it will also be useful for LV and MV, so it should be made a generic machine.

crazyR14:
@ShadowNinja it was put into `machines/other` as its a admin only node. just like an anchor is. if you think it belongs in `machines/hv` then feel free to move it upon merging.

and you are quite right it would be usefull LV and MV and that is the plan for the future as well as making the EU rate configurable which is another reason its been placed into `machines/other`

I dont paticulaly have the time at the moment to add the above features. but in its current form its been tested heavilly on a live server.

ShadowNinja:
Machines in the `{L,M,H}V/` folders aren't restricted to being usable by non-admins; they're only supposed to be filtered by their tier.  The anchor is in `other/` because it doesn't use technic's electrical system.

HybridDog:
my mod adds an admin power node, which is connectible to lv, mv and hv:
https://github.com/HybridDog/technic_extras/commit/96d29b8c35b7188efc53670a76dab1012ee02295

l didn't make it possible to change the power it gives because l thought it isn't necessary.

VanessaE:
No further response from the author.  Closing...

PR #224, opened 2015-04-11
Add Vacuum cleaner

Phvli:
Adds a power tool that picks up entities from the ground. Live-tested on a public survival server for several months. Proved to be useful with chainsaws, and the crafting price seems to be about right now, too.


HybridDog:
please add sounds

Phvli:
Good point, it lacks those. I'll try to wrap something up soon.

VanessaE:
What is the current status of this pull request?

Phvli:
Due to a theft, I've got no recording equipment at the moment. Anyone with a suitable sound effect is welcome to add it. If nothing else turns up, I will record something this fall but it may take a while.

HybridDog:
there are free vacuum cleaner sounds:
http://freesound.org/search/?q=vacuum+cleaner

lf you want to download a noise there you don't need to register/login, you can also download the demo sound: (firefox) load the site, press ctrl+shift+q, click on the play button and see the link appearing on the bottom of the firefox window

RealBadAngel:
Merged with sound for the tool (and slightly increased radius)

PR #228, opened 2015-06-06
default:mese_block is now default:mese

asl97:
https://github.com/minetest-technic/technic/blob/master/technic/crafts.lua#L189


VanessaE:
7cd848a050b053c469efd20d89c7e2b34b6005ce

PR #229, opened 2015-06-09
Fix technic_worldgen incompatible with snow

asl97:
see https://github.com/Splizard/minetest-mod-snow/issues/14


HybridDog:
l think its fixed the other way round.

PR #231, opened 2015-07-09
fixed crash when HV quarry mines bones (dead player)

petermaloney:
In thread 7176328ba780:
/home/peter/projects/manjaro/minetest/src/minetest-0.4.12/src/server.cpp:505: void Server::step(float): A fatal error occurred: ...sr/share/minetest/games/minetest/mods/bones/init.lua:46: attempt to index local 'player' (a nil value)
stack traceback:
        ...sr/share/minetest/games/minetest/mods/bones/init.lua:46: in function 'can_dig'
        ...minetest/mods/technic/technic/machines/HV/quarry.lua:153: in function 'technic_run'
        .../mods/technic/technic/machines/switching_station.lua:225: in function 'run_nodes'
        .../mods/technic/technic/machines/switching_station.lua:231: in function <.../mods/technic/technic/machines/switching_station.lua:177>

Issue #172, opened 2015-01-07
Quarry prone to crashing.


kaeza:
The quarry calls the node's `on_dig` callback passing `nil` as `player`.

This confuses mods which expect `player` to never be `nil`, like `doors` and `signs_lib`.

The MT Lua API does not mention in the documentation for any of the relevant functions that the player can be a nil value, so it is a bug on the part of the technic mod.

To reproduce:
* Put a door or a sign from [signs_lib](https://github.com/VanessaE/homedecor_modpack). I'm sure there are tons of other nodes out there for which this bug also manifests.
* Put a quarry next to it.
* Power up the quarry to dig the node placed in step 1.

Stack traceback: https://gist.github.com/kaeza/c629bd768b09f4edfca7

It was discussed on the IRC channel, and the best approach would be to implement it using "fake player refs".

Leaving this here so it does not get lost.

ShadowNinja:
The code originally used `dig_node`, which has some extra checks and calls callbacks.  It should be changed back.

It also uses a whole bunch of completely uncommented math now that's hard to understand.

t4im:
In fact switching back to `dig_node` will _continue_ calling the can_dig with a nil player.
So that is currently official behavior.

As mentioned in irc, this is formost an issue to be solved in minetest first.
Either the documentation is missing this information about nil-players being allowed to be passed
or alternativly minetest requires a change of `dig_node`

Then again with minetest/minetest#2015 dig_node might need a small behavioral overhaul anyway.

VanessaE:
I haven't seen quarries crash in quite a long time, let alone from anything referenced here.

t4im:
And there's [smoketest](https://github.com/t4im/shakedown#smoketest) now to find exactly this kind of issues (test for this particular case included) :)

PR #232, opened 2015-07-10
handling exceptions when quarry digs (eg. when player is nil)

petermaloney:
issues 172, 231, for example with item bones:bones

(sorry I have no idea how to tell github it's issue 172 or 231 again)

This should handle the general case, as an exception handler rather than as a "bug fix" (since the bug is in the other code that assumes player is not nil)


petermaloney:
actually I'm not sure what happens with the "can_dig = false" lines in the function... (unlike in every other language I used with try/catch). I don't think they have any effect any more, going out of scope before they are used later, but didn't notice anything in testing. What should they do? It doesn't get stuck trying to dig air, and doesn't trigger the exception code for non-diggable things I know (like chests with things in it)... so I don't really know what it does.

asl97:
:-1: this isn't really the way to do it, i suggest making an issue at https://github.com/minetest/minetest_game since the api has specify that player is optional. 
related: #172 #231

petermaloney:
So I did some testing, and I think actually this anonymous function does change the can_dig, so it works fine.

And how is handling exceptions not one option of a right way to do it? It doesn't fix the root cause, but that's the whole point of "exceptions". This is making the code fail gracefully instead of crashing the whole server. And unlike my previous quick fix, this one handles the general case in this one block of code. (are there others we should handle too?)

There are plenty of people out there that think bones works fine, and then when they add the quary, and see a crash, they immediately blame technic and not bones. This will handle it better than blame shifting will. We blame them; they blame you; nothing gets done... right?

asl97:
downstream shouldn't have to suffer just cause upstream doesn't know or want to fix the bug due to not following the api.

what if other mod also use on_dig on the bones?  it would crash.  should every single mod use exception just cause of it and slow down everything, making a messy nested code?

I don't think so.  upstream should fix it and it is just a simple pull request or issue to let the people upstream know about it.

the fix is just a simple one liner `if not player then return end` to add in the bones mod instead of this messy nested stuff.

i didn't make the issue or pull request cause i wasn't the one who gotten the error.

t4im:
The api defines this only after minetest/minetest#2108 which was created due to exactly this kind of confusion. Many mods, including the upstream ones, might simply not have catched up with the decision yet, but should be fixed no matter of a workaround here or elsewhere. Opening up a upstream bugreport would be the sensible thing to do either way.

Ekdohibs:
Closing, as there are problems with rebasing and this should anyway be fixed by https://github.com/minetest/minetest_game/pull/875

PR #235, opened 2015-07-25
minor changes rubber tree leaves

wischi-chr:
+ added a description for rubber tree leaves
+ added them to the leaves group (so they can be used in generic crafting recipes)
+ removed not_in_creative_inventory because (imo) there is no good reason to remove them


asl97:
@wischi-chr 
oh lol, you could have just rebase but since another pull request is already make, just leave this as close.
____

note to others: new pull request at https://github.com/minetest-technic/technic/pull/237

wischi-chr:
oh f*ck sry

PR #238, opened 2015-07-29
Use moreblocks as an optional dependency.

1Nick1Name:
See: https://forum.minetest.net/viewtopic.php?f=11&t=2538&start=975#p185712


VanessaE:
cfd788358cc81c568d3aa295d2fd69b761f8ce83

PR #243, opened 2015-08-20
Allow mods to change damage caused by radiation

HybridDog:
https://github.com/minetest-technic/technic/pull/243/files?w=1


stujones11:
I would also like to see this merged, my 3d_armor mod is now ready to make good use of it.

https://github.com/stujones11/minetest-3d_armor/tree/HybridDog-lead_armor

HybridDog:
What can l do that l get informed automatically if someone adds a label?

HybridDog:
rebased

PR #247, opened 2015-09-05
made lasers 10x more precise

petermaloney:


asl97:
wouldn't this make it 10x heavier?

maybe undoing the removal of the "antialiasing" by https://github.com/minetest-technic/technic/pull/159 would be better than this.

related:
https://github.com/minetest-technic/technic/issues/163
https://github.com/minetest-technic/technic/pull/159
maybe [#100 by this comment](https://github.com/minetest-technic/technic/pull/100#issuecomment-55533537)

ShadowNinja:
This results in checking the same node ten times, which is way too much, it also seems odd that this would increase accuracy.

VanessaE:
This doesn't appear to be desired.  Sorry, @HybridDog

HybridDog:
Sorry, github shows references again after rebasing.

PR #249, opened 2015-09-12
frames: use functions from vector.extras to improve speed, play sounds when digging their objects and fix #170

### LINKED ISSUE #170: frames cheat

HybridDog:


technic/machines/other/frames.lua
```diff
@@ -411,23 +419,25 @@ minetest.register_entity("technic:frame_entity", {
...
-		local pointed_thing = {type = "node", under = pos_under, above = pos_above}
+		local pointed_thing = {type = "node", under = pos, above = pos_above}
+
 		local itemstack = clicker:get_wielded_item()
 		local itemdef = minetest.registered_items[itemstack:get_name()]
-		if itemdef ~= nil then
+		if itemdef then
 			itemdef.on_place(itemstack, clicker, pointed_thing)
 		end
+		clicker:set_wielded_item(itemstack)
```
HybridDog: https://github.com/minetest-technic/technic/issues/170

technic/machines/other/frames.lua
```diff
@@ -200,129 +314,39 @@ local nodeboxes= {
...
-			else
-				--local pointed_thing = {type = "node", under = pos}
-				if pointed_thing then
-					minetest.item_place_node(itemstack, placer, pointed_thing)
-				end
-			end
+		on_rightclick = function(...)
+			return rightclick_frame(...)
 		end,
 	})
```
HybridDog: wrapping the functions should disable making a new one for every frame node

ShadowNinja:
Perhaps instead of iterating over a list checking every entry and doing three table lookups on each, you could use a hash lookup.  You can use `minetest.get_pos_hash` or `minetest.pos_to_string` to generate a unique hash that you can use as an index to check if a position is in the table.

technic/machines/other/frames.lua
```diff
@@ -697,6 +718,18 @@ local function compress_templates(pos)
...
+	local t = {}
+	local function helper(line) table.insert(t, line) return "" end
+	helper((str:gsub("(.-)\r?\n", helper)))
+	return t
+end
+
+local function pos_from_string(str)
+	local l = lines(str)
+	return {x = tonumber(l[1]), y = tonumber(l[2]), z = tonumber(l[3])}
+end
```
ShadowNinja: Er, why not just `minetest.string_to_pos`?

technic/machines/other/frames.lua
```diff
@@ -664,6 +678,13 @@ local function expand_template(pos)
 	end
 end
 
+local function pos_to_string(pos)
+	if pos.x == 0 then pos.x = 0 end -- Fix for signed 0
+	if pos.y == 0 then pos.y = 0 end -- Fix for signed 0
+	if pos.z == 0 then pos.z = 0 end -- Fix for signed 0
+	return tostring(pos.x).."\n"..tostring(pos.y).."\n"..tostring(pos.z)
+end
```
ShadowNinja: Why not `minetest.pos_to_string`?

technic/machines/other/frames.lua
```diff
@@ -664,6 +678,13 @@ local function expand_template(pos)
 	end
 end
 
+local function pos_to_string(pos)
+	if pos.x == 0 then pos.x = 0 end -- Fix for signed 0
+	if pos.y == 0 then pos.y = 0 end -- Fix for signed 0
+	if pos.z == 0 then pos.z = 0 end -- Fix for signed 0
```
ShadowNinja: What is this for, isn't it pointless?

technic/machines/other/frames.lua
```diff
@@ -470,53 +487,59 @@ mesecon.register_mvps_unmov("technic:damage_entity")
...
+		local nodename = get_frame(pos1) or minetest.get_node(pos1).name
+		if not pos_in_list(c,pos1)
+		and nodename~="air"
+		and (
+			minetest.registered_nodes[nodename].frames_can_connect == nil
+			or minetest.registered_nodes[nodename].frames_can_connect(pos1,vect)
+		) then
+			c[#(c)+1] = pos1
+			if minetest.registered_nodes[nodename].frame == 1 then
+				local adj = minetest.registered_nodes[nodename].frame_connect_all(nodename)
```
t4im: The last four `minetest.registered_nodes[nodename]` uses will likely crash when trying to move connected unknown_nodes.
They also could have been looked up once into a local variable (this is in a loop after all!)

technic/machines/other/frames.lua
t4im: From the lua_api: "^ if defined, itemstack will hold clicker's wielded item"
So this looks a bit redundant. Can you explain, why this would make any difference for #170?

technic/machines/other/frames.lua
```diff
@@ -200,129 +314,39 @@ local nodeboxes= {
...
-			l1={{x=-1,y=0,z=0},{x=1,y=0,z=0},{x=0,y=-1,z=0},{x=0,y=1,z=0},{x=0,y=0,z=-1},{x=0,y=0,z=1}}
-			for i,dir in ipairs(l1) do
-				if string.sub(nodename,-7+i,-7+i)=="1" then
-					l2[#(l2)+1]=dir
-				end
-			end
-			return l2
+		groups = groups,
+		frame_connect_all = function(nodename)
+			return connect_frame(nodename)
```
t4im: Wrapping the functions additionally to moving them out of the loop is redundant:

Just use:
```lua
frame_connect_all = connect_frame,
...
on_rightclick = rightclick_frame,
```

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+		obj = minetest.add_entity(pos, "technic:frame_entity")
+		obj:get_luaentity():set_node({name=node.name})
+
+		return itemstack
+	else
+		--local pointed_thing = {type = "node", under = pos}
+		if pointed_thing then
+			minetest.item_place_node(itemstack, placer, pointed_thing)
+		end
+	end
```
t4im: Does not return the leftover itemstack.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+		end
+
+		obj = minetest.add_entity(pos, "technic:frame_entity")
+		obj:get_luaentity():set_node({name=node.name})
+
+		return itemstack
+	else
+		--local pointed_thing = {type = "node", under = pos}
+		if pointed_thing then
+			minetest.item_place_node(itemstack, placer, pointed_thing)
```
t4im: Does not return the returned leftover itemstack.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+	if minetest.is_protected(pos, pname) then
+		minetest.log("action", pname
+			.. " tried to place " .. nodename
+			.. " at protected position "
+			.. minetest.pos_to_string(pos))
+		minetest.record_protection_violation(pos, pname)
+		return itemstack
+	end
+	local node = minetest.get_node(pos)
+	if node.name == "air" then
```
t4im: Check for buildable_to instead.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+	node.name = nodename
+	minetest.set_node(pos,node)
+end
+
+local function place_frame(itemstack, placer, pointed_thing)
+	local pos = pointed_thing.above
+	if not pos then
+		return
+	end
+	local pname = placer:get_player_name()
```
t4im: Should now check for buildable_to, content_ignore and unknown nodes and decide based on that to perhaps choose pointed_thing below, before doing the protection check.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+	local nodename = itemstack:get_name()
+	if minetest.is_protected(pos, pname) then
```
t4im: Should now check for buildable_to, content_ignore and unknown nodes and decide based on that to perhaps choose pointed_thing below, before doing the protection check.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+	minetest.set_node(pos,node)
+end
+
+local function place_frame(itemstack, placer, pointed_thing)
+	local pos = pointed_thing.above
+	if not pos then
+		return
+	end
+	local pname = placer:get_player_name()
+	local nodename = itemstack:get_name()
```
t4im: Should now check for buildable_to, content_ignore and unknown nodes and decide based on that to perhaps choose pointed_thing below, before doing the protection check.

technic/machines/other/frames.lua
t4im: Better check buildable_to instead of air.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+		nodename = string.sub(nodename,1,-2)..newstate
+	end
+	node.name = nodename
+	minetest.set_node(pos,node)
+end
+
+local function place_frame(itemstack, placer, pointed_thing)
+	local pos = pointed_thing.above
+	if not pos then
+		return
```
t4im: This does not return the leftover itemstack.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
 end
 
 local function is_supported_node(name)
-	return ((string.find(name, "tube") ~= nil) and (string.find(name, "pipeworks") ~= nil))
+	return string.find(name, "tube") ~= nil and string.find(name, "pipeworks") ~= nil
```
t4im: Tubes have special definitions in their node definition to check for.
The name is not a good thing to rely on and might end in some false positives and false negatives alike.

technic/machines/other/frames.lua
```diff
@@ -91,19 +89,31 @@ local function add_table(table,toadd)
...
 		end
 	end
+	dones = nil
 	for _,pos in ipairs(poslist) do
-		local npos=vector.add(pos,vect)
+		local npos = vector.add(pos,vect)
 		local name = minetest.get_node(npos).name
-		if ((name~="air" and minetest.registered_nodes[name].liquidtype=="none") or frames_pos[pos_to_string(npos)]) and not(pos_in_list(poslist,npos)) then
+		if (
+			(name ~= "air" and minetest.registered_nodes[name].liquidtype == "none")
```
t4im: It appears you want buildable_to here.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+				.. " at protected position "
+				.. minetest.pos_to_string(pos))
+			minetest.record_protection_violation(pos, pname)
+			return itemstack
+		end
+
+		minetest.set_node(pos, {name = nodename})
+
+		local take_item = true
+		local def = minetest.registered_items[nodename]
```
t4im: Will crash if right clicked with an unknown node.
Do we even need to check items here? Doesn't registered_nodes suffice?

technic/machines/other/frames.lua
```diff
@@ -578,7 +591,8 @@ local function template_connected(pos,c,connectors)
 	for _,vect in ipairs({{x=0,y=1,z=0},{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=-1,z=0}}) do
 		local pos1=vector.add(pos,vect)
 		local nodename=minetest.get_node(pos1).name
-		if not(pos_in_list(c,pos1)) and (nodename=="technic:template" or nodename == "technic:template_connector")then
+		if not(pos_in_list(c,pos1))
```
t4im: `pos_in_list` only has one return value, so it's not needed to throw away the "others" (i.e. using extra brackets).

technic/machines/other/frames.lua
```diff
@@ -200,129 +314,39 @@ local nodeboxes= {
 	if zm==0 then
 		table.insert(nodeboxes, {-b,-b,-a, b,b,-b})
 	end
-	
-	local nameext=tostring(xm)..tostring(xp)..tostring(ym)..tostring(yp)..tostring(zm)..tostring(zp)
+
+	local nameext = tostring(xm)..tostring(xp)..tostring(ym)..tostring(yp)..tostring(zm)..tostring(zp)
```
t4im: Better use `string.format` here: it avoids all these the concatenations and `tostring()`s (note, that in 5.1 tostring can only be skipped for numbers) and makes it generally easier to read.

technic/machines/other/frames.lua
```diff
@@ -149,11 +157,117 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
...
+			minetest.record_protection_violation(pos, pname)
+			return itemstack
+		end
+
+		minetest.set_node(pos, {name = nodename})
+
+		local take_item = true
+		local def = minetest.registered_items[nodename]
+		-- Run callback
+		if def.after_place_node then
```
t4im: Will crash if right clicked with an unknown node.

technic/machines/other/frames.lua
t4im:
```lua
print(tostring(-1.0 * 0))
```
will yield `-0`.

technic/machines/other/frames.lua
HybridDog: l didn't change the code yet.

technic/machines/other/frames.lua
HybridDog: fixed

technic/machines/other/frames.lua
HybridDog: This tells if it's a node which can be put inside a frame, l think, not only tubes should be able to become put there.

technic/machines/other/frames.lua
HybridDog: > check for buildable_to

no, frames may contain a buildable_to cable from some mod.
>content_ignore and unknown nodes

it aborts if it's not air or not a node which can be put inside
>perhaps choose pointed_thing.under

l don't think it's necessary.

technic/machines/other/frames.lua
HybridDog: see ↑

technic/machines/other/frames.lua
HybridDog: l think if it's a node, it should be placed via a given function.

technic/machines/other/frames.lua
HybridDog: it can't be an unknown one

technic/machines/other/frames.lua
HybridDog: fixed (untested)

technic/machines/other/frames.lua
t4im: What i meant here: you don't need to do a lookup on the larger `registered_items` table, when a lookup on `registered_nodes` is enough (smaller table). If it's not a node, then the `set_node` would already have done some awkward stuff anyway by now.

HybridDog: The set_node isn't safe enough l think, idk if the on_place gets executed

technic/machines/other/frames.lua
HybridDog: fixed

technic/machines/other/frames.lua
t4im: It's a convention for all `on_place` methods. Things become awkward without it, when trying to place it e.g. onto a flower, grass or into water. Perhaps think of another airlike node, that would make it impossible to place a frame then, like moontest for example.

technic/machines/other/frames.lua
HybridDog: idk exactly how this works, if l just use that, the function becomes copied for every node and more memory is used, isn't it?

technic/machines/other/frames.lua
t4im: Of course it can. It just needs to pass the `is_supported_node test`, which is atm just a matter of the right name. I've seen servers disable certain pipeworks tubes for various reasons (there is a config file for that). These tubes became unknown nodes with the right name for this scenario.

technic/machines/other/frames.lua
HybridDog: this shouldn't use simple recursion which could cause stack overflow

technic/machines/other/frames.lua
t4im: Lua is one of the languages that passes references, not values:
from http://www.lua.org/manual/5.1/manual.html#2.2
> Tables, functions, threads, and (full) userdata values are objects: variables do not actually contain these values, only references to them. Assignment, parameter passing, and function returns always manipulate references to such values; these operations do not imply any kind of copy. 

The only thing that is "copied" is the reference, which is usually not much more than an address (often a simple integer). Wrapping just adds another level of nesting here.

technic/machines/other/frames.lua
t4im: `on_place` is called on the wielding item when the user right clicks with it (which usually means you want to place it somewhere)
That's also why you usually test if `pointed_thing` has an `on_right_click`: If it has, you would not place the node, but instead call the pointed_thing's `on_right_click` instead and return the itemstack.

`set_node` internally only causes `on_construct` of the newly placed node to be called.

technic/machines/other/frames.lua
t4im: That is true, too. Tail calls might help.

HybridDog: fixed

technic/machines/other/frames.lua
HybridDog: ok, thanks

technic/machines/other/frames.lua
HybridDog: ok, fixed

technic/machines/other/frames.lua
HybridDog: l need to update the is_supported_node test

technic/machines/other/frames.lua
HybridDog: idk, l didn't edit the templates

technic/machines/other/frames.lua
HybridDog: see above

technic/machines/other/frames.lua
HybridDog: set_node is bad if you want to simulate placing

technic/machines/other/frames.lua
HybridDog: idk, l didn't edit the templates

technic/machines/other/frames.lua
Ekdohibs: This code was needed to make sure pos_to_string returns the same result for two equal vectors, even if there are signed zeros in them as they are used as hash keys. Maybe minetest.hash_node_position could work, however there is a problem with the serialization of these hashes IIRC (there was an issue about it some time ago).

PR #250, opened 2015-10-02
Change tile_images to tiles.

MT-Modder:
For mods that search for the tiles field.


ShadowNinja:
2d86ee64e360e7a2825bbd7c46ac3d40596d39ab

PR #258, opened 2016-02-28
Zinc is too rare!!! It is an issue!

mrsteyk:
Changed zinc parameters in first block block to match "lead noise" and "tin frequent". Please commit this! Zinc was a real problem to me untill i changed this file, now i can find about 10 zinc in 2 minutes!

### DIFF
--- a/technic_worldgen/oregen.lua
+++ b/technic_worldgen/oregen.lua
@@ -54,13 +54,13 @@ minetest.register_ore({
 	ore_type         = "scatter",
 	ore              = "technic:mineral_zinc",
 	wherein          = "default:stone",
-	clust_scarcity   = 8*8*8,
-	clust_num_ores   = 4,
-	clust_size       = 3,
+	clust_scarcity   = 9*9*9,
+	clust_num_ores   = 5,
+	clust_size       = 7,
 	y_min       = -32,
 	y_max       = 2,
-	noise_params     = zinc_params,
-	noise_threshhold = zinc_threshhold,
+	noise_params     = lead_params,
+	noise_threshhold = lead_threshhold,
 })
 
 minetest.register_ore({


Ekdohibs:
All technic ores generate in a non-uniform distribution, to avoid the rainbow cave effect and to make mining more interesting. However, this could be made optional, see #259

amianthoidal:
I think the better solution is to disable the non-uniform generation (as discussed in #259 ). If you go into /technic_worldgen and open oregen.lua, comment out (using '--' at the start of the line) all lines that start with 'noise_params' (maybe 'noise_threshold' too to be safe).

I've been experimenting with solutions for a few hours and this is what has worked best. I have seen other posts suggesting to copy over the lead parameters, but this can still leave you stuck without zinc if you find yourself in a dead-zone on the map. If you do find some I think it's more random chance than anything. On some maps the solution works, but on others I can't find any zinc by no-clipping. On the other hand, disabling the non-uniform generation works like a charm.

It's strange that I've seen more complaints about zinc than other minerals. Personally, I've found lead to be extremely common in my survival game. I've also found some chromium, marble, and granite. Is there maybe a sort of bias with the noise having to do with its initial value? I also find that snow biomes are extremely common in my games.

numberZero:
The player should have choice to enable or disable the non-uniform distribution, as proposed in #259. But zinc should be fixed anyway, see #274.

numberZero:
@VanessaE why have you merged this?

numberZero:
This PR looks broken. It uses `lead_params` for zinc, that’s certainly wrong: seed, at least, should be different for different ores. It also uses `lead_threshold` for zinc instead of decreasing `zinc_threshold`; this way, this 1. is incorrect, and 2. affects upper (-32..2) layer of the map only.
Just a small decrease of `zinc_threshold` could solve the problem that zinc may be very hard to find for first time, as it is not rare in map as a whole, it is just distributed very non-uniformly.

VanessaE:
I was under the impression that it was good.  *shrug*

mrsteyk:
Found out after so many days its was merged. IDK how it worked time when i was trying that but now it doesnt work as it used to be. _shrug_ So i think it was placebo or something.

mrsteyk:
I can agree with numberZero this PR made it only more rare. What really helps is deleting the noise paramaters in zinc registration!

Issue #246, opened 2015-09-03
Make blast resistant concrete resists dungeon master blasts


pyrollo:
Is it possible to add Blast Resistant Concrete node to the unbreakable group ? 

technic/concrete/init.lua line 94 :

- 	groups={cracky=1, level=3, concrete=1},
+ 	groups={cracky=1, level=3, concrete=1, unbreakable=1},

So blast resistant concrete wont be destroy by tenplus1 dungeon masters.

t4im:
It _is_ breakable though. But it would probably make sense overwriting [on_blast](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L3251), provided @tenplus1 makes use of that in the mod (which i couldn't find right now).

Ekdohibs:
This should be in on_blast indeed, to make it resistant to tnt too.

MT-Modder:
The mobs PR: https://github.com/tenplus1/mobs/pull/47 has been merged. Maybe now the above linked PR can be merged as well?

ShadowNinja:
0c144cd75c638f1f1c67ae91973f3ff8ffc781dd

PR #260, opened 2016-02-28
Make blast-resistant concrete blast resistant

MT-Modder:
Helps with #246
~~I'll be checking https://github.com/tenplus1/mobs for on_blast checks.~~

Mobs PR: https://github.com/tenplus1/mobs/pull/47 Merged!

### DIFF
--- a/concrete/init.lua
+++ b/concrete/init.lua
@@ -99,6 +99,7 @@ minetest.register_node(":technic:blast_resistant_concrete", {
 	after_dig_node = function (pos, oldnode, oldmetadata, digger)
 		technic.update_posts(pos, false)
 	end,
+	on_blast = function() end,
 })
 
 minetest.register_node(":technic:concrete_post_platform", {
--- a/concrete/init.lua
+++ b/concrete/init.lua
@@ -99,7 +99,12 @@ minetest.register_node(":technic:blast_resistant_concrete", {
 	after_dig_node = function (pos, oldnode, oldmetadata, digger)
 		technic.update_posts(pos, false)
 	end,
-	on_blast = function() end,
+	on_blast = function(pos, intensity)
+		if intensity > 1 then
+			minetest.remove_node(pos)
+			minetest.add_item(pos, "technic:blast_resistant_concrete")
+		end
+	end,
 })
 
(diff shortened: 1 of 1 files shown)


ShadowNinja:
This should check the intensity, so that more intense explosions will destroy it.

MT-Modder:
A blast of tnt uses an intensity of 1:
https://github.com/minetest/minetest_game/blob/master/mods/tnt/init.lua#L96

Does this look good?
```
on_blast = function(pos, intensity)
    if intensity > 1 then
        minetest.remove_node(pos)
    end
end,
```

ShadowNinja:
According to `lua_api.txt` 1 is mid range of regular TNT (although it isn't actually calculated by the mod yet).  I think it should be a bit more resistant than that.  Perhaps 1.5 or so.  Other than that, yes, that looks good.

MT-Modder:
Okay.

It's probably safe to just use the nodename itself instead of fetching it from ```minetest.get_node(pos).name``` since it's within the nodedef, right?

ShadowNinja:
Yes.  You should return the item instead of dropping it though -- as soon as the docs API gets updated to support that.
I'll merge this soon.

ShadowNinja:
0c144cd75c638f1f1c67ae91973f3ff8ffc781dd

MT-Modder:
Okay thanks I'll submit a PR for that once https://github.com/minetest/minetest_game/pull/862 is merged.

It should look like this right?
```
on_blast = function(pos, intensity)
    if intensity > 1 then
        minetest.remove_node(pos)
        return ItemStack("technic:blast_resistant_concrete")
    end
end,
```

sofar:
check the doors mod for how to do this:

e.g. ```return { "technic:blast_resistant_concrete" }```

MT-Modder:
Ah thank you very much. Pretty simple. I'll submit the PR now since it's already supported. I hadn't realized.

MT-Modder:
After this https://github.com/minetest/minetest_game/pull/862 PR the blast intensity next to TNT is about 9 and the intensity at the edge of the blast is about 1. We should probably adjust the check for intensity otherwise it isn't really blast-resistant.

After it's merged of course.

PR #266, opened 2016-03-20
on_blast() changes for new TNT api

MT-Modder:
1. Return nodename in on_blast() callback
2. Re-tune intensity check for blast-resistant concrete

### DIFF
--- a/concrete/init.lua
+++ b/concrete/init.lua
@@ -86,7 +86,7 @@ minetest.register_node(":technic:blast_resistant_concrete", {
 	on_blast = function(pos, intensity)
 		if intensity > 1 then
 			minetest.remove_node(pos)
-			minetest.add_item(pos, "technic:blast_resistant_concrete")
+			return {"technic:blast_resistant_concrete"}
 		end
 	end,
 })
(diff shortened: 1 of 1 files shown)


ShadowNinja:
To merge once this is actually supported and documented.

MT-Modder:
Supported since: https://github.com/minetest/minetest_game/commit/12d5ca2f48694c990658d401e5f68d281f260aee
Documented since: https://github.com/minetest/minetest_game/commit/098ea0d10a181ea1bcc98d72766969b5c94cd945

MT-Modder:
Oh and according to: https://github.com/minetest/minetest_game/commit/ccee025ce3e3f56596aeda69f8dc32940bba7b57

Intensity is always 1.0 or higher so I'll do some testing and then open a new PR.

MT-Modder:
Okay so after testing for awhile I found that checking for an intensity that's 9 or below results in node removal, when using one TNT right next to it.

So with the new check it takes three TNT next to it to destroy it. Making it fairly blast-resistant.

MT-Modder:
@ShadowNinja Could you test this to see if it's to your liking?

Remember checking for anything lower than 9 blows up the concrete with one TNT (directly next to it). At 9 it starts at three TNT needed to blow it up (All three need to be directly next to it).

MT-Modder:
Bump

MT-Modder:
@ShadowNinja Any news for this?

PR #267, opened 2016-03-22
Radiation changes

ShadowNinja:
Changes:
  * Switch reactor to a lead shield (SS was just used as a placeholder since lead wasn't available at the them).  Old reactors converted.
  * Adjust radioactivity math.  This removes the 1000 and 0.25 multipliers, making the math a bit simpler.  It also makes radiation resistance more effective (a large part of the protection from extra layers was just the extra distance).
  * Split radiation code from nuclear reactor.  Radiation isn't necessarily tied to the reactor.
  * Add longer-term radiation damage.  I reduced radioactivity of various things to account for the increased damage.


MT-Modder:
:+1: Nice!

MT-Modder:
So was this reverted or something?

ShadowNinja:
Erm, it looks like someone removed them with a force-push.

ShadowNinja:
Re-pushed.

MT-Modder:
Sweet!

PR #268, opened 2016-03-22
Mining laser improvements

### LINKED ISSUE #163: Mining laser's path doesn't match the aim point
I haven't tracked down when the change occurred but the current mining laser path of travel appears to originate from lower than the aim point and so the aim is off.

ShadowNinja:
Changes:
- Adjust start offset (it's actually closer to 1.6)
- Update particle API usage (vel/acc -> velocity/acceleration).
- Replace `trace_node_ray` algorithm (old one would sometimes go through the wrong node).
- Make the mining laser use a "fat" beam.  This and the previous should fix #163.

### DIFF
--- a/technic/tools/mining_lasers.lua
+++ b/technic/tools/mining_lasers.lua
@@ -38,8 +38,8 @@ local function laser_node(pos, node, player)
 		minetest.remove_node(pos)
 		minetest.add_particle({
 			pos = pos,
-			vel = {x=0, y=2, z=0},
-			acc = {x=0, y=-1, z=0},
+			velocity = {x=0, y=2, z=0},
+			acceleration = {x=0, y=-1, z=0},
 			expirationtime = 1.5,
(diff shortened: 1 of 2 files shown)


technic/tools/mining_lasers.lua
```diff
@@ -61,17 +61,17 @@ local function laser_shoot(player, range, particle_texture, sound)
 
 	local start_pos = vector.new(player_pos)
 	-- Adjust to head height
-	start_pos.y = start_pos.y + 1.9
+	start_pos.y = start_pos.y + 1.6
```
HybridDog: The camera position is 1.625 above the player position if you ignore bobbing.
https://github.com/minetest-technic/technic/compare/master...HybridDog:hversion#diff-be93baf033504fc8693772842af598c6R122

t4im:
Very nice. This works noticeably better. :+1:

PR #271, opened 2016-03-29
Fix for issue 270

NathanSalapat:
I think I may have messed up with the pull request, as it shows more commits than it should.
Anyway this fixes the crashing when trying to dig a machine or battery that has upgrades.


Ekdohibs:
:+1:

Ekdohibs:
#273 was merged instead.
https://github.com/minetest-technic/technic/commit/322615542e570f6363c7006b90f334e9ffd722c8

PR #273, opened 2016-04-06
Fix item drop when digging mv furnace with upgrade

tenplus1:
Fixed crash when digging mv furnace containing upgrade


Ekdohibs:
:+1:

Ekdohibs:
Note: this and #271 fix the same bug in a different way: this one drops the items (which looks like what was intended), while the other forbids digging the machine. The question now is: which one do we want to merge?

numberZero:
In my opinion, items should be dropped (as when digging node breaker, for example: it drops the tool contained, if any). Moreover, I think it should be applied to chests too. But as some players won’t agree with me, it can be made configurable.

tenplus1:
Definitely drop upgrades within, this will fix machines from crashing for now and additional settings can be added at a later date.

Ekdohibs:
https://github.com/minetest-technic/technic/commit/322615542e570f6363c7006b90f334e9ffd722c8

PR #275, opened 2016-04-10
Configurable radiation

numberZero:
This pull request contains several modifications:
1. Entities can be hurt by radiation (may be disabled in the config)
2. Long-term radiation effects may be disabled
3. Radiation protection is supported (using the “radiation” armor group instead of callbacks proposed earlier); this may be disabled too (see also [my pull request](https://github.com/stujones11/minetest-3d_armor/pull/53) to the 3d_armor mod)
4. Shielding is sane again (line 269→271 in technic/radiation.lua; the change I reverted looks accidental, as it makes a block of dirt be almost enough to protect from corium radiation)

### DIFF
--- a/technic/radiation.lua
+++ b/technic/radiation.lua
@@ -257,14 +257,12 @@ local function apply_fractional_damage(o, dmg)
 	return false
 end
 
-local function dmg_player(pos, player, strength)
-	local pl_pos = player:getpos()
-	pl_pos.y = pl_pos.y + abdomen_offset
+local function calculate_base_damage(node_pos, object_pos, strength)
 	local shielding = 0
-	local dist = vector.distance(pos, pl_pos)
+	local dist = vector.distance(node_pos, object_pos)
 
-	for ray_pos in technic.trace_node_ray(pos,
-			vector.direction(pos, pl_pos), dist) do
+	for ray_pos in technic.trace_node_ray(node_pos,
+			vector.direction(node_pos, object_pos), dist) do
 		local shield_name = minetest.get_node(ray_pos).name
 		shielding = shielding + node_radiation_resistance(shield_name) * 0.1
 	end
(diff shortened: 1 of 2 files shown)


technic/radiation.lua
```diff
@@ -257,26 +261,69 @@ local function apply_fractional_damage(o, dmg)
...
+local function calculate_damage_multiplier(object)
+	local ag = object.get_armor_groups and object:get_armor_groups()
+	if not ag then
+		return 0
+	end
+	if ag.immortal then
+		return 0
+	end
+	if ag.radiation then
+		return 0.01 * ag.radiation
```
ShadowNinja: Please explain this multiplier, it looks arbitrary.

ShadowNinja:
I made shielding more effective on purpose.  A dirt node still shouldn't offer full protection from a corium source though.

The rest looks good.

technic/radiation.lua
numberZero: `calculate_damage_multiplier` does the following:

1. It checks is the object damageable at all.
2. If the object has the `radiation` armor group, it uses its value. As with default damage with the `fleshy` AG, its value is not protection but percent of damage actually done.
3. If the object have no `radiation` AG but have `fleshy` one, it is still considered to be damageable by radiation; that’s necessary to support existing mods. But as its value varies a lot, I take a square root to reduce variance.

Actually that function is mostly a workaround for current mods that don’t support radiation yet. It seems reasonable that entities that are “`fleshy`” are vulnerable to radiation too, but the level may only be guessed.

numberZero:
Shielding actually needs careful tweaking as it is a bit too weak by default, although in realty meters (of concrete) _are_ required to protect from radiation.

It’s a bit difficult to make reasonable coefficients in simplified game mechanics as radiation resistance against different types of radiation differs a lot; moreover, some materials that protect well e.g. from gamma-rays (like lead) offer little protection from neutrons, _and vice versa_ (for example, concrete protects well from neutrons but not from X-rays)

numberZero:
So may it be merged?
If you expect me to e.g. correct shielding coefficient I will do necessary testing and changes.

Also it may make sense to make overall multipliers configurable (both for damage and shielding).

ShadowNinja:
I'll merge this -- once I get to it...

Fixer-007:
Bump

ShadowNinja:
1810f417d2028f84dbed31038edf674cbfc50e1c, 51be33deea1c6ed942f18ba4f02580540541d2c9, b739ed6cb1e24eb2800a4b99222b927acfbef9d0, 73afc40d9c9125c2e07137422569c30150e0d84c, cbe97434dc6524cb8fc5163936e726db095ab222

Issue #285, opened 2016-06-24
enable/disable converter


DanHolli123:
there should be an option for enabling and disabling the power converters or disable the 10000 discharge if not used

numberZero:
Yes, just as in force field emitter (that can be controlled manually or by mesecon signal). That would be much more handy than using pistons.

DanHolli123:
or without mesecons just right clicking

numberZero:
Created corresponding pull request.

@DanHolli123 if running own server, you can check out [my updated version](https://github.com/numberZero/technic/tree/control-supply_converter) right now (the `control-supply_converter` branch)

DanHolli123:
It's for personal use but I'll give it a try!

VanessaE:
Fixed with #288

Issue #168, opened 2014-12-10
Digilines connectivity


cheapie:
It would be great if there was a way to monitor power supply and demand using Jeija's "digilines" mod. I do have something like this working here as a demonstration: https://github.com/cheapie/technic_digilines_interface

However, my solution is not optimal. I think that a better way to do this (above my skill level, unfortunately) would be to have a separate "interface" or "monitor" node, possibly as a separate mod in the Technic modpack.

I was talking with VanessaE about this, and she had the idea that the switching station could be made to do this (like I did) but using "Upgrade Slots" like the MV machines do. Technic could be made to soft-depend on digilines, and if digilines is present, certain nodes such as the switching station could have (for example) Luacontrollers placed in the upgrade slots to give them digilines communication capabilities.

DS-Minetest:
Suggestion: Use the power monitor as digiline distributor.

HybridDog:
A mesecons signal could instigate sending batbox info.

DS-Minetest:
> batbox

Ehm, I don't think, you mean this: https://en.wikipedia.org/wiki/BatBox

Make the power monitor only send changes if it gets a mesecon signal and send info if it gets "GET" command via digiline?

DS-Minetest:
Next step battery box: https://github.com/DS-Minetest/technic/tree/battery_box-digilines

DS-Minetest:
I've planned to add some interaction with digilines for:
- [x] switching station
- [ ] battery box (PR already started)
- [ ] nuclear reactor (with digiline_remote)
- [ ] forcefield generator
- [ ] quarry

DS-Minetest:
I've decided not to do anything with quarry since that makes no sense.

numberZero:
> I've decided not to do anything with quarry since that makes no sense.

Well, just imagined a robot that places quarries, configures them using digilines, waits for them to finish, digs and places in other place... ugh...

Seriously, I may name one thing that does make sense: a quarry might report when it finishes digging. Also it might report when it touches something non-diggable. It doesn’t need to be controlled via digilines, though.

DS-Minetest:
You are totally right with both points. I'll re-add quarry to the list.

PR #288, opened 2016-07-23
Supply converter control

numberZero:
Supply converter certainly needs some way to control it (see #285, at least). Full-featured digiline control would be better, of course, but needs more effort.

Note: most of the code I added I copied from the Forcefield generator code.

### DIFF
--- a/technic/machines/supply_converter.lua
+++ b/technic/machines/supply_converter.lua
@@ -10,26 +10,36 @@
 local S = technic.getter
 
 local function set_supply_converter_formspec(meta)
-	local formspec = "size[5,1.5]"
+	local formspec = "size[5,2.25]"..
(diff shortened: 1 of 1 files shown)


ForbiddenJ:
Oh man! I've been using this cobbled-together type of setup for a long time to control these converters. It includes a piston, wire, and cable.

screenshot from 2017-04-01 23-05-52

DS-Minetest:
Hm, I never saw this commit before. Maybe because some servers have outdated technic or because I haven't done much with converters since some time or because I accepted that the piston was the only way.

> Full-featured digiline control would be better, of course, but needs more effort.

I'll add this to my list in #168.

numberZero:
@ForbiddenJ I did that too) and a node detector watching battery charge level, yes ;)

PR #290, opened 2016-07-25
nuclear reactor edit

harrierjack:
Reactor core can be attached to tubes and sorts rods in equal stacks

I made it for myself (it's too dangerous manually handling fuel rods ;) ) for a fully auto nuke.

### DIFF
--- a/technic/machines/HV/nuclear_reactor.lua
+++ b/technic/machines/HV/nuclear_reactor.lua
@@ -111,7 +111,7 @@ section through the middle:
 	CCCC CCCC
 	CBBB BBBC
 	CBSS SSBC
-	CBSWWWSBC
+	CBSWTWSBC
 	CBSW#WSBC
 	CBSW|WSBC
 	CBSS|SSBC
@@ -119,6 +119,7 @@ section through the middle:
 	CCCC|CCCC
 	C = Concrete, B = Blast-resistant concrete, S = Stainless Steel,
 	W = water node, # = reactor core, | = HV cable
+	T = tube (or water)
 
 The man-hole and the HV cable are only in the middle, and the man-hole
 is optional.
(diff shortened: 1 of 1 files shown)


t4im:
Not sure it needs refueling that often to really need to automate it. :-)
You know, you can always use nyan cats to protect you, when handling the rods:
technic_nyancat_protection

This works, since it allows 2 holes in the stainless steel. One for the cable, other for the cat to "take the heat". Selection box is reachable from there.

I 'm not a big fan of solving things programmatic, that can already be solved by creativity in-game alone and well. That kinda takes the challenges out of the game.

If you solve one challenge, you should replace it with another one, imo.
Automation? Fine. But then increase the rod consumption so much, that you'd actually need it.
Want to have a water node less there? Maybe enforce some special type of active water cooling system.

<sub>no cats were harmed in making this screenshot. :smile_cat:  <sub>

harrierjack:
It's not about how often, its about making somewhat of a functional nuclear plant, one where you don't go to the danger zone carrying fuel rods in your pocket :)

Also i'm not a big fan of solving things with magical cat blocks so at least with this you can be creative in making your power plant, though granted, if you build another plant for some  nodebreaker/switcher logic it's not impossible just gets big and complicated very fast, and while we have tubes, might as well use em.

tbh, I originally thought of making something of a submerged tube counting as water block but couldnt figure out a good way of doing so. And since the core needed to be edited slightly to be tube-enabled, slightly altering the core reactor was the easiest solution.

At first I just had an override 'mod' only overriding core reactor with this implementation, but then figured that probably ppl liking technic stuff might want to automate things so figured I'd send in the small changes. 

Feel free to do take it or leave it :)

VanessaE:
Looks like the consensus is "leave it".  Sorry @harrierjack (and sorry again for the uncustomary delay).

PR #301, opened 2016-10-24
Added line to manual about how to enable windmills

cyian-1756:


ShadowNinja:
This is incorrect, you're not supposed to change `config.lua`.  The change is supposed to occur in the world's `technic.conf`.

VanessaE:
No word from the author.  closing...

PR #304, opened 2016-12-07
Make admin anchor work with pistons

infmagic2047:


VanessaE:
What is the intended purpose of this pull request, exactly?  I'm not supportive of the idea of being able to push an anchor around....

numberZero:
`mesecon.register_mvps_stopper("technic:admin_anchor")`
That would be better.

infmagic2047:
Well, I have built a mining machine that will move by itself, and I want to keep it loaded. Maybe it is better to fix this in mesecons and pipeworks, so feel free to close this pr if you want.

PR #306, opened 2017-01-15
Update has_locked_chest_privilege to current minetest_game function

Thomas--S:


numberZero:
Wow. Should work, but looks weird for me.

PR #308, opened 2017-01-30
Make mk3 drill in 3x3 mode drill nodes from top to bottom

thatgraemeguy:
The mk3 mining drill, when operating in 3x3 mode, first drills the horizontal row being pointed at, then the row above, then the row below. When the top row contains falling nodes, they end up not being drilled, as they fall after the middle row has been drilled.

Before: https://www.youtube.com/watch?v=GuS-UVGAlSk

After: https://www.youtube.com/watch?v=eir_1wRPuMw


HybridDog:
The mining laser should have the same problem.

PR #310, opened 2017-02-26
Pass info to called technic_run at which stage we are.

agaran:
This lets to save cpu/lua cycles for PR/RE machines like supply converters. It does not break ones not using info, but patching (same way as supply convert is patched in this PR), saves half of execution time on not running same code over same data TWICE.

### DIFF
--- a/technic/machines/supply_converter.lua
+++ b/technic/machines/supply_converter.lua
@@ -13,7 +13,7 @@ local run = function(pos, node, run_name)
 	local demand = 10000
 	local remain = 0.9
 
-	if run_name == technic.receive then 
+	if run_name == technic.receiver then 
 		-- do not run TWICE (it is PR-RE machine), save lua cycles
 		return 
 	end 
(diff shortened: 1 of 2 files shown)


technic/machines/supply_converter.lua
```diff
@@ -9,9 +9,12 @@
 
 local S = technic.getter
 
-local run = function(pos, node)
+local run = function(pos, node, run_name)
 	local demand = 10000
 	local remain = 0.9
+
+	if run_name == "RE" then return end -- do not run TWICE (it is PR-RE machine), save lua cycles
```
SmallJoker: Please expand this to three (+ comment) lines.

ShadowNinja:
You should use the constants from `register.lua` (`technic.producer` etc).

VanessaE:
Needs rebased also.

agaran:
I'll make new one then someday.

PR #323, opened 2017-03-15
Fix supply converter bug

### LINKED ISSUE #322: new supply coverter is broken
When it is enabled, it converts up to its rating put in field, when it gets disabled by button, it converts 10kEU instead of being off.

numberZero:
Should fix #322 
@agaran please test


VanessaE:
7d5329834dd749eca2bcc5d07c5cc016bd743a38

PR #329, opened 2017-03-23
make the switching station able to send supply and demand via digiline

DS-Minetest:
related: #168 
Give a switching station a mesecons signal and it will always send demand and supply (in a table) via digiline (with channel `"swtiching_station"..minetest.pos_to_string(pos)` where pos is the position of the switching station) when demand or supply changes.
Send via digiline "GET" or "get" and the switching station will send you on the same channel supply and demand back.

### DIFF
--- a/technic/machines/switching_station.lua
+++ b/technic/machines/switching_station.lua
@@ -63,6 +63,23 @@ minetest.register_node("technic:switching_station",{
 		pos.y = pos.y - 1
 		minetest.forceload_free_block(pos)
 	end,
+	mesecons = {effector = {
+		rules = mesecon.rules.default,
+	}},
+	digiline = {
+		receptor = {action = function() end},
+		effector = {
+			action = function(pos, node, channel, msg)
+				if msg == "GET" or msg == "get" then
+					local meta = minetest.get_meta(pos)
+					digilines.receptor_send(pos, digilines.rules.default, channel, {
+						supply = meta:get_int("supply"),
+						demand = meta:get_int("demand")
+					})
+				end
+			end
+		},
+	},
 })
 
 --------------------------------------------------
@@ -190,7 +207,7 @@ minetest.register_abm({
 		local BA_nodes
 		local RE_nodes
 		local machine_name = S("Switching Station")
-		
+
 		-- Which kind of network are we on:
 		pos1 = {x=pos.x, y=pos.y-1, z=pos.z}
 
@@ -216,7 +233,7 @@ minetest.register_abm({
 			minetest.forceload_free_block(pos1)
 			return
 		end
-		
+
 		-- Run all the nodes
 		local function run_nodes(list, run_stage)
 			for _, pos2 in ipairs(list) do
@@ -231,7 +248,7 @@ minetest.register_abm({
 				end
 			end
 		end
-		
+
 		run_nodes(PR_nodes, technic.producer)
 		run_nodes(RE_nodes, technic.receiver)
 		run_nodes(BA_nodes, technic.battery)
@@ -301,6 +318,20 @@ minetest.register_abm({
 				S("@1. Supply: @2 Demand: @3",
 				machine_name, technic.pretty_num(PR_eu_supply), technic.pretty_num(RE_eu_demand)))
 
+		-- digiline
+		if minetest.get_modpath("mesecons") and
+				minetest.get_modpath("digilines") and
+				mesecon.is_powered(pos) then
+			if PR_eu_supply ~= meta:get_int("supply") or
+					RE_eu_demand ~= meta:get_int("demand") then
+				digilines.receptor_send(pos, digilines.rules.default,
+						"switching_station"..minetest.pos_to_string(pos), {
+							supply = PR_eu_supply,
+							demand = RE_eu_demand
+						})
+			end
+		end
+
 		-- Data that will be used by the power monitor
 		meta:set_int("supply",PR_eu_supply)
 		meta:set_int("demand",RE_eu_demand)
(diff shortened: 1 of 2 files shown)


DS-Minetest:
Still no comment. :confused:

technic/machines/switching_station.lua
```diff
@@ -301,6 +318,20 @@ minetest.register_abm({
 				S("@1. Supply: @2 Demand: @3",
 				machine_name, technic.pretty_num(PR_eu_supply), technic.pretty_num(RE_eu_demand)))
 
+		-- digiline
+		if minetest.get_modpath("mesecons") and
+				minetest.get_modpath("digilines") and
```
numberZero: cache that

numberZero: cache that

technic/machines/switching_station.lua
```diff
@@ -301,6 +318,20 @@ minetest.register_abm({
...
+				mesecon.is_powered(pos) then
+			if PR_eu_supply ~= meta:get_int("supply") or
+					RE_eu_demand ~= meta:get_int("demand") then
+				digilines.receptor_send(pos, digilines.rules.default,
+						"switching_station"..minetest.pos_to_string(pos), {
```
numberZero: channel should be configurable

technic/machines/switching_station.lua
```diff
@@ -63,6 +63,23 @@ minetest.register_node("technic:switching_station",{
...
 		minetest.forceload_free_block(pos)
 	end,
+	mesecons = {effector = {
+		rules = mesecon.rules.default,
+	}},
+	digiline = {
+		receptor = {action = function() end},
+		effector = {
+			action = function(pos, node, channel, msg)
+				if msg == "GET" or msg == "get" then
```
numberZero: no check for channel. imagine you have several switching stations (e.g. MV and HV) connected to a single digiline wire.

MT-Modder:
I honestly can't wait! You've tested it (extensively)?

technic/machines/switching_station.lua
```diff
@@ -301,6 +318,20 @@ minetest.register_abm({
...
 
+		-- digiline
+		if minetest.get_modpath("mesecons") and
+				minetest.get_modpath("digilines") and
+				mesecon.is_powered(pos) then
+			if PR_eu_supply ~= meta:get_int("supply") or
+					RE_eu_demand ~= meta:get_int("demand") then
+				digilines.receptor_send(pos, digilines.rules.default,
+						"switching_station"..minetest.pos_to_string(pos), {
+							supply = PR_eu_supply,
```
numberZero: 7th indentation level? is that really necessary?

technic/machines/switching_station.lua
DS-Minetest: So again use `"swtiching_station"..minetest.pos_to_string(pos)` as channel?

DS-Minetest:
@MT-Modder How often will you write this?

technic/machines/switching_station.lua
DS-Minetest: I think so.

DS-Minetest:
I need to do more testing.

numberZero:
er, I meant to cache the boolean value, e.g.
`local use_digilines = minetest.get_modpath("mesecons") and minetest.get_modpath("digilines") and true`
although this way it is readable too, so not a problem.
(btw, as you use soft-dependency, you can use smth like `minetest.global_exists('mesecon')` instead of `get_modpath`. but I am not sure what is better)

numberZero:
so... waiting for this to be merged =) that would be a great thing for automation)
@DS-Minetest I wanted to ask you to do the same for battery boxes (as another PR, of course) but noticed you made this PR from `master`. Don’t do that next time, at least because it prevents you from making several PRs at once. Use feature branches. (ex.: `git checkout -b digiline-switching-station`)

DS-Minetest:
Yes, I'm very unexperienced with git, I always do it with master and I should look some time how to do it better.

Perhaps I should check instead for the global `digilines` and `mesecons` variables.

Note, this is still nearly not tested, I don't even know, if it shuts down if mesecons or digilines is not installed or if different channels work.

kaeza:
It seems the channel is not being saved?

I don't have much experience with actual use of digilines, so can somebody show an example setup for testing?

numberZero:
oh well, it needs to handle `on_receive_fields`...

kaeza:
@numberZero Thanks! It seems to work. Will add the `on_receive_fields` myself and merge it.

kaeza:
https://github.com/minetest-technic/technic/commit/6abd857b3fc0dab74c71d77a26b05e412d43f562

Thank you for the contribution!

DS-Minetest:
> it needs to handle `on_receive_fields`

Yep, that was the thing that was missing, I would have added it after testing.

This was a very fast merge, I hope there are no bugs.

> Perhaps I should check instead for the global digilines and mesecons variables.

I still don't know, would it be better than checking for modpath? I mean checking for a variable should be lighter than using a function for search for the modpath.

Todo:
> do the same for battery boxes

DS-Minetest:
_If anyone wants to know (probably not),_ I've started to make something similar to battery boxes: https://github.com/DS-Minetest/technic/tree/battery_box-digilines
A PR will be started when I've finished it.

@numberZero thanks for the information about the `git checkout -b` thing!

kaeza:
>I still don't know, would it be better than checking for modpath? I mean checking for a variable should be lighter than using a function for search for the modpath.

That is run at load time, so doesn't matter.

>Yep, that was the thing that was missing, I would have added it after testing.
>
>This was a very fast merge, I hope there are no bugs.

I tested with #0's code and it worked as I would expect. If somebody discovers bugs, they can be fixed. It's not like the codebase is now magically locked :)

It's fine to do devices in separate commits. Makes it easier to revert individual changes if something goes wrong.

PR #335, opened 2017-03-29
Battery box digilines

DS-Minetest:
Send useful information via digilines at get command and on run if mesecons signal is on and it gets input.
This time it's tested, so I've also fixed the crashes in switching station.
related: #168 (#329)

### DIFF
--- a/technic/machines/register/battery_box.lua
+++ b/technic/machines/register/battery_box.lua
@@ -64,17 +64,17 @@ function technic.register_battery_box(data)
 	local formspec =
 		"size[8,9]"..
 		"image[1,1;1,2;technic_power_meter_bg.png]"..
-		"list[current_name;src;3,1;1,1;]"..
+		"list[context;src;3,1;1,1;]"..
 		"image[4,1;1,1;technic_battery_reload.png]"..
-		"list[current_name;dst;5,1;1,1;]"..
+		"list[context;dst;5,1;1,1;]"..
 		"label[0,0;"..S("%s Battery Box"):format(tier).."]"..
 		"label[3,0;"..S("Charge").."]"..
 		"label[5,0;"..S("Discharge").."]"..
 		"label[1,3;"..S("Power level").."]"..
 		"list[current_player;main;0,5;8,4;]"..
-		"listring[current_name;dst]"..
+		"listring[context;dst]"..
 		"listring[current_player;main]"..
-		"listring[current_name;src]"..
+		"listring[context;src]"..
 		"listring[current_player;main]"
 	if digilines then
 		formspec = formspec..
@@ -83,12 +83,12 @@ function technic.register_battery_box(data)
 
 	if data.upgrade then
 		formspec = formspec..
-			"list[current_name;upgrade1;3.5,3;1,1;]"..
-			"list[current_name;upgrade2;4.5,3;1,1;]"..
+			"list[context;upgrade1;3.5,3;1,1;]"..
+			"list[context;upgrade2;4.5,3;1,1;]"..
 			"label[3.5,4;"..S("Upgrade Slots").."]"..
-			"listring[current_name;upgrade1]"..
+			"listring[context;upgrade1]"..
 			"listring[current_player;main]"..
-			"listring[current_name;upgrade2]"..
+			"listring[context;upgrade2]"..
 			"listring[current_player;main]"
 	end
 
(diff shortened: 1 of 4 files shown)


technic/machines/switching_station.lua
```diff
@@ -48,6 +48,15 @@ minetest.register_craft({
 	}
 })
 
+local mesecon_def
+if mesecons_path then
+	mesecon_def = {effector = {
+		rules = mesecon.rules.default,
+	}}
+else
+	mesecon_def = nil
```
numberZero: redundant, variables are `nil` by default. that’s not C

technic/machines/register/battery_box.lua
```diff
@@ -98,7 +105,25 @@ function technic.register_battery_box(data)
 			EU_upgrade, tube_upgrade = technic.handle_machine_upgrades(meta)
 		end
 		local max_charge = data.max_charge * (1 + EU_upgrade / 10)
-			
+
+		-- send digiline information
+		if mesecons_path and digilines_path and mesecon.is_powered(pos) then
+			if eu_input ~= 0 then
```
numberZero: why do you check this? a batbox can change charge level even without external power, by (dis)charging tools

technic/machines/register/battery_box.lua
```diff
@@ -206,7 +242,46 @@ function technic.register_battery_box(data)
...
+							demand = meta:get_int(tier.."_EU_demand"),
+							supply = meta:get_int(tier.."_EU_supply"),
+							input  = meta:get_int(tier.."_EU_input"),
+							charge = meta:get_int("internal_EU_charge"),
+							max_charge = data.max_charge * (1 + technic.handle_machine_upgrades(meta) / 10),
+							src      = inv:get_stack("src", 1):to_table(),
+							dst      = inv:get_stack("dst", 1):to_table(),
+							upgrade1 = inv:get_stack("upgrade1", 1):to_table(),
+							upgrade2 = inv:get_stack("upgrade2", 1):to_table()
+						})
```
numberZero: the message becomes larger for batboxes... what’s about moving its construction to a separate function that may be called both from here and from `run`?

technic/machines/register/battery_box.lua
DS-Minetest: >_< Forgot about that. But then it would always send every second would that be good?

technic/machines/register/battery_box.lua
DS-Minetest: I'm not sure if this would be better since in the run function there are already given some variables and so, I don't have to get them extra again.

technic/machines/register/battery_box.lua
```diff
@@ -62,29 +65,33 @@ function technic.register_battery_box(data)
...
 		"list[current_player;main;0,5;8,4;]"..
-		"listring[current_name;dst]"..
+		"listring[context;dst]"..
 		"listring[current_player;main]"..
-		"listring[current_name;src]"..
+		"listring[context;src]"..
 		"listring[current_player;main]"
+	if digilines_path then
+		formspec = formspec..
+			"field[0.28,4.1;3.2,1;channel;Digiline Channel;${channel}]"
```
Thomas--S: How about adding a submit button? (e.g. for mobile users)

technic/machines/register/battery_box.lua
numberZero: That’s only if it receives mesecon signal, so not that bad. Although I don’t think that’s needed at all, it may send data on request only. But it should also send data when something happens, e.g. when a tool comes in or out.

technic/machines/register/battery_box.lua
DS-Minetest: Don't they have a submit/enter button on their touch keyboard?

Perhaps it would be just better to make a edit channel button since:
> Problematically, the channel is always deleted if you haven't pressed enter and the battery box gets input.

technic/machines/register/battery_box.lua
DS-Minetest: So, shall I delete the part in run?
I did it anyway only for consistence with switching station.

technic/machines/register/battery_box.lua
numberZero: you should check for channel to be non-nil (i.e. `if fields.channel then...` as only `nil` evaluates to `false`) instead: you get `""` if the field is empty and the player presses enter, but you may get `nil` at any time, regardless of the field value.

DS-Minetest:
Ok, this changes should fix all comments, I think.

numberZero:
d141052 works well for me :+1:
b20af69 works too but is worse IMO.

kaeza:
b20af69 sounds reasonable to me to work around the bug, and I would've done the same. It can be changed later if needed.

You have my :+1:.

DS-Minetest:
> work around the bug

Well, not for a real bug. It was only inconvenient. Also like this android players have no problems.

kaeza:
I tested it and seems to work as expected.

Merged in https://github.com/minetest-technic/technic/commit/338f3b6a997bb3ebea9d62d986b7c3327f06072e.

Thank you!

PR #336, opened 2017-03-29
Update mg.lua

numberZero:


kaeza:
I would have appreciated a more meaningful commit message :(

PR #338, opened 2017-04-01
make forcefield controlabe with digilines

DS-Minetest:
Commands are `get`, `range <int>`, `off`, `on`, `toggle`, `shape <0/1>` and everyone again with capital letters.
related: #168

### DIFF
--- a/technic/machines/HV/forcefield.lua
+++ b/technic/machines/HV/forcefield.lua
@@ -164,12 +164,12 @@ local digiline_def = {
 	receptor = {action = function() end},
 	effector = {
 		action = function(pos, node, channel, msg)
-			msg = msg:lower()
 			local meta = minetest.get_meta(pos)
+			if channel ~= meta:get_string("channel") then
+				return
+			end
+			msg = msg:lower()
 			if msg == "get" then
-				if channel ~= meta:get_string("channel") then
-					return
-				end
 				digilines.receptor_send(pos, digilines.rules.default, channel, {
 					enabled = meta:get_int("enabled"),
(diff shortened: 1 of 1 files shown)


technic/machines/HV/forcefield.lua
```diff
@@ -150,6 +160,40 @@ local mesecons = {
...
+					range   = meta:get_int("range"),
+					shape   = meta:get_int("shape")
+				})
+			elseif msg == "off" then
+				meta:set_int("enabled", 0)
+			elseif msg == "on" then
+				meta:set_int("enabled", 1)
+			elseif msg == "toggle" then
+				local onn = meta:get_int("enabled")
+				onn = -(onn-1) -- switch 0 to 1 and vice versa
```
DS-Minetest: `onn = 1-onn`

DS-Minetest:
Everything is ok now IMO. Could please someone review?

kaeza:
https://github.com/minetest-technic/technic/commit/ef8bb38bfd8e52a511fbe930257b173ddab93f63

Thanks!

numberZero:
@DS-Minetest thank for your work, but

> `range <int>`

please don’t use such homegrown serialization formats. Digilines do support tables, you know. Only simplest commands, like `get`, may be sent as plain strings.

Also... what if `msg` is not a string? `msg:lower()` will crash the server then, I suppose.

DS-Minetest:
I made the range thing to string so, that you can easily type it by hand. Both could be allowed.

For the type of msg should really be checked. Perhaps I'll open a new PR to improve this..

Issue #292, opened 2016-07-29
Use minetest.clear_craft to remove recipes


Wuzzy2:
Look at here:
https://github.com/minetest-technic/technic/blob/master/technic/crafts.lua#L166

Minetest has now the awesome function `minetest.clear_craft` to remove crafting recipes. It should be used everywhere instead of this little hack.

MT-Modder:
:+1:

VanessaE:
69feeef0afc0d66dd8809481c434ba0afa33600e

DS-Minetest:
Hm, I think, I didn't get all override to 0 recipes eg. colors.

PR #339, opened 2017-04-01
use clear_craft

DS-Minetest:
related: #292

### DIFF
--- a/technic/crafts.lua
+++ b/technic/crafts.lua
@@ -1,6 +1,23 @@
 -- check if we have the necessary dependencies to allow actually using these materials in the crafts
 local mesecons_materials = minetest.get_modpath("mesecons_materials")
 
+-- Remove some recipes
+-- Bronze
+minetest.clear_craft({
+	type = "shapeless",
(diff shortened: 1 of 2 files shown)


Thomas--S:
I highly doubt that this works as expected.

According to [`lua_api.txt`](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L2045), you can specify either output or input only.

DS-Minetest:
Oh, you are right, so now input is ignored. Shall input or output be cleared?

Thomas--S:
I think that depends on two things:

- Are the recipes cleared before the new ones are registered?
- Do we want to
   - disable one specific recipe? => clear input
   - disable all recipes except the ones that the technic mod defines for a certain item? => clear output

DS-Minetest:
Yep, we want to clear it before registering the others, so I moved up. Also we want to clear the outputs, so that the technic recipes are the only valid ones. (So, the moreores tin-copper recipe for bronze is also cleared now, I think, that's good.)

PR #341, opened 2017-04-02
digilinize nuclear_reactor

DS-Minetest:
related: #168 
My digiline_remote mod (https://github.com/DS-Minetest/digiline_remote) is used to communicate with the reactor core.
New formspec:
screenshot_20170402_170900


thatgraemeguy:
Obvious ones I can think of:

- Get percentage remaining of fuel currently being used
- Get count of unused fuel rods per slot

Can't think that there's much else. I would not add on/off commands, I recall it was specifically stated that a reactor is not meant to be switched off once it has started running.

DS-Minetest:
Thanks @thatgraemeguy.
- The percentage can already be gotten if you know the maximal value, which is static.
- Count of unused fuel rods is a good idea.
- Perhaps send something when the hull gets damaged, so some sort of alarm signal (and also when it stops).
- Would a self-destruct be command good? :smile: 
- I thought of editing the nuce reactor so, that you have to press a button to start the process, I'll also add a digiline command for that.

DS-Minetest:
Ok, now there are some pretty useful features.

numberZero:
> Perhaps send something when the hull gets damaged, so some sort of alarm signal (and also when it stops).

Mandatory.

Actually there are little useful features currently available. I may suggest some sort of power level control; IRL reactors have that, although some can’t be stopped reversibly (but they need continuous fuel supply as well).

And remember: `msg` may not be a string. Check its type, e.g. `if type(msg) ~= "string" then return end` (string comparison is very fast in Lua)

technic/machines/HV/nuclear_reactor.lua
```diff
@@ -231,28 +270,6 @@ local function run(pos, node)
 	local burn_time = meta:get_int("burn_time") or 0
 
 	if burn_time >= burn_ticks or burn_time == 0 then
```
numberZero: So it stops after burning one fuel set now, right? Not sure is that OK. Perhaps it should continue operation automatically (IIUC, just remove `burn_time == 0` that is used for “cold start”), or there should be an option.

technic/machines/HV/nuclear_reactor.lua
```diff
@@ -13,6 +13,9 @@ local burn_ticks = 7 * 24 * 60 * 60  -- Seconds
 local power_supply = 100000  -- EUs
 local fuel_type = "technic:uranium_fuel"  -- The reactor burns this
 
+local digiline_meltdown = minetest.setting_get("technic_nuce_digiline_selfdestruct") == "true"
```
DS-Minetest: Actually there's another setting system with technic as it seems (eg. for wind mill).

technic/machines/HV/nuclear_reactor.lua
```diff
@@ -268,6 +285,66 @@ local function run(pos, node)
 	end
 end
 
+local nuclear_reactor_receive_fields = function(pos, formname, fields, sender)
+	local meta = minetest.get_meta(pos)
+	if fields.remote_channel then
+		meta:set_string("remote_channel", fields.remote_channel)
```
DS-Minetest: There was never any protection test for digiline.
The channel should be secret.

DS-Minetest:
Thanks for comments but this is still
> **WIP**

. (Perhaps this means here `Work In Pause`.) I'll continue improving this PR when I got time and co..

technic/machines/HV/nuclear_reactor.lua
DS-Minetest: I'll add a possibility (in formspec) to disable digiline for a nuclear reactor completely.

DS-Minetest: Done.

DS-Minetest:
Ok, this is now longer WIP.

technic/machines/HV/nuclear_reactor.lua
numberZero: > possibility to disable digiline

Very good, it’s rather dangerous) But that’s not what I meant: currently anyone who can right-click on the reactor can re-enable digiline support, change the digiline channel, etc. That’s what need to be fixed.

numberZero:
Looks good) But [add protection support](https://github.com/minetest-mods/technic/pull/341/files/042f860100272fe12299aa5567a923cb0fa78bc5..f9549d2f9466a6c55e152f82a43a25ea12e5ff9e#r114061961) (not sure why that comment is not listed on this page)

technic/machines/HV/nuclear_reactor.lua
DS-Minetest: Hm, true.
Is is also possible to steal fuel rods?

DS-Minetest: Done.

DS-Minetest:
I also added missing `on_receive_fields` to active core.

PR #343, opened 2017-04-03
add cable plates

DS-Minetest:
Feel free to suggest a better name.
screenshot_20170403_201106
They can also conduct power (in all directions):
screenshot_20170403_201125
Practical use case:
screenshot_20170403_201321
You could also build a wall where you can easily place a machine to it without that it looks meh.
Code is very messy.

### DIFF
--- a/technic/machines/register/cables.lua
+++ b/technic/machines/register/cables.lua
@@ -215,6 +215,7 @@ function technic.register_cable(tier, size)
 			def.groups.not_in_creative_inventory = 1
 		end
 		minetest.register_node("technic:"..ltier.."_cable_plate_"..i, def)
+		cable_tier["technic:"..ltier.."_cable_plate_"..i] = tier
 	end
 
 	local c = "technic:"..ltier.."_cable"
--- a/technic/machines/register/cables.lua
+++ b/technic/machines/register/cables.lua
@@ -121,7 +121,8 @@ function technic.register_cable(tier, size)
 	local ltier = string.lower(tier)
 	cable_tier["technic:"..ltier.."_cable"] = tier
 
-	local groups = {snappy=2, choppy=2, oddly_breakable_by_hand=2}
+	local groups = {snappy=2, choppy=2, oddly_breakable_by_hand=2,
+			["technic_"..ltier.."_cable"] = 1}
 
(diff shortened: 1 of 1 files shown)


thatgraemeguy:
Existing cables already conduct in all directions, as long as there is another machine/cable in that direction. I don't really understand what this is?

VanessaE:
It looks to me like it's meant to solve the issue where a machine might be smaller than a cube.  In such a case, one of these objects looks like it would serve to "plug the hole" that would be left under the machine with a regular cable.

DS-Minetest:
This is only thought to look good. It conducts the same way as normal cables do.

VanessaE:
Messy code or not, it has my :+1: as long as it works as intended.

kaeza:
I give a :+1: as well, but would like this:

>I think I'd prefer the majority of the texture to look fairly bland (e.g. steel plating, or perhaps plastic), with only the central part textured so that it looks like a cable sticking into/through the plate.

kaeza:
Merged as https://github.com/minetest-technic/technic/commit/4dda59da0ce714712749c56f71e36c2de75ee2d2. Somebody should provide suitable textures if desired.

Thank you!

DS-Minetest:
This really got merged and so fast. O_o
The textures can't be changed that well I think, since it's a nodebox and so, it could look bad.
I had planned to use fine_pointed if the placer presses aux1 on place. But because of the daily ppa problems I didn't yet. Perhaps later.

kaeza:
@DS-Minetest The way the nodebox is done allows to make such a texture. See for example the back texture for the fencepost-attached sign in signs_lib:

signs_post_back

The code is not magically locked; if you want to add a feature or fix something feel free to submit patches.

VanessaE:
Besides, perhaps it could be made into a mesh node if texturing is too difficult.

DS-Minetest:
> Besides, perhaps it could be made into a mesh node if texturing is too difficult.

But mesh can't be connected.

You could make a + shape with another texture. I didn't do that to reduce texture file amount.

VanessaE:
Good point, I forget that cables themselves are connected nodeboxes.

PR #345, opened 2017-04-09
Take Item from Cable Plate Stack on Place

Thomas--S:
~_untested_~

### DIFF
--- a/technic/machines/register/cables.lua
+++ b/technic/machines/register/cables.lua
@@ -182,7 +182,7 @@ function technic.register_cable(tier, size)
 			tiles = {"technic_"..ltier.."_cable.png"},
 			groups = table.copy(groups),
 			sounds = default.node_sound_wood_defaults(),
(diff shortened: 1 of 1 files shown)


technic/machines/register/cables.lua
```diff
@@ -210,6 +210,8 @@ function technic.register_cable(tier, size)
 					end
 				end
 				minetest.set_node(pointed_thing.above, {name = "technic:"..ltier.."_cable_plate_"..num})
+				itemstack:take_item()
```
VanessaE: Need to check for creative mode (or priv) before doing that.

Thomas--S:
I hope this works like expected. :wink:

Thomas--S:
Tested now, should work like expected.

Thomas--S:
Changed the drop to a recipe.

Tested again, everything works fine.

@VanessaE This is ready for merge in my opinion.

Thomas--S:
Thanks for merging!

PR #346, opened 2017-04-13
Fix fence nodebox

Thomas--S:
See minetest/minetest#5572


DS-Minetest:
Now I also know, why some connections were not pointable.

PR #347, opened 2017-04-13
Generalize check for cable below battery box

Thomas--S:
Because we have cable plates now, too.


VanessaE:
Force-pushed this out of the repo, didn't work right.

PR #348, opened 2017-04-13
Generalize check for cable below battery box

Thomas--S:
This time better tested and should really work. Sorry for the inconvenience caused by the first PR.


Thomas--S:
Thanks for merging!

PR #354, opened 2017-04-15
light_source at maximum value is 14

CarlBarr:
corrected warning


DS-Minetest:
Why not use `default.LIGHT_MAX`?

PR #357, opened 2017-04-22
advanced cable plate placing

DS-Minetest:
raster
If you hold the use key / aux1 / turbo / E, you can place it into another direction.

### DIFF
--- a/technic/depends.txt
+++ b/technic/depends.txt
@@ -2,6 +2,7 @@ default
 pipeworks
 technic_worldgen
 bucket?
+screwdriver?
 mesecons?
 mesecons_mvps?
 digilines?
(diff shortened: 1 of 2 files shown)


VanessaE:
Ordinarily, shift/sneak is used to initiate an alternate placement or whatever.  Might want to consider it.

DS-Minetest:
Yes but if you fly you don't want to fly down while placing precisely. Also for what else would you need to press the aux1 button while placing a cable plate?
So, I think, aux1 makes more sense than sneak.

VanessaE:
Well, for me, aux1 IS the fly-down key (an assignment I kept from before shift became the default).

DS-Minetest:
I'll make both possible as compromise.

DS-Minetest:
Now you have to hold one of them but not both.

DS-Minetest:
I hope, this comment now doesn't extend the range of this PR.

DS-Minetest:
Could someone please review this PR? It's lying here already for weeks ready to be merged.

PR #358, opened 2017-04-24
make solar arrays only work with sunlight

DS-Minetest:
This uses `param1` instead of `minetest.get_node_light(pos1, nil)`.
Also this allows to have solar arrays beyond 0, but that's no real problem since they are there very inefficient.
I don't know, why this hasn't been done yet.

### DIFF
--- a/technic/machines/register/solar_array.lua
+++ b/technic/machines/register/solar_array.lua
@@ -1,17 +1,17 @@
 
 local S = technic.getter
 
+local function get_light(param1)
+	local alight = (param1-param1%16)/16
+	local sunlight = (param1%16 ~= alight and param1%16) or 0
+	return sunlight, alight
+end
+
 function technic.register_solar_array(data)
 	local tier = data.tier
 	local ltier = string.lower(tier)
 
-	local run = function(pos, node)
-		-- The action here is to make the solar array produce power
-		-- Power is dependent on the light level and the height above ground
-		-- There are many ways to cheat by using other light sources like lamps.
-		-- As there is no way to determine if light is sunlight that is just a shame.
-		-- To take care of some of it solar panels do not work outside daylight hours or if
-		-- built below 0m
+	local function run(pos, node)
 		local pos1 = {}
 		local machine_name = S("Arrayed Solar %s Generator"):format(tier)
 		pos1.y = pos.y + 1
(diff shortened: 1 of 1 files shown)


Ekdohibs:
`param1` only stores daylight and nightlight, not sunlight and daylight for now, so this does not work properly. Instead, we are waiting on this change to come to the engine before changing it.

DS-Minetest:
You can get the current light with time of day. This PR doesn't do this yet, so on morning and evening the supply won't become smaller but will just stop at once. I don't see the problem here.
@Ekdohibs I really don't know, what you mean, it works very well.

HybridDog:
https://github.com/HybridDog/technic/blob/hversion/technic/machines/register/solar_array_new.lua

numberZero:
@DS-Minetest 
Place a solar panel under a mese lamp (or 3×3 roof of mese lamps) at +100. It won’t work, I suppose. But with stone node instead of lamps, it will work.

DS-Minetest:
Hm, does air have param1?

HybridDog:
yes

DS-Minetest:
Then I'll check the air above as it was before.

DS-Minetest:
Ok, now the only remaining problem is that solar arrays on a hight can only be on or off. But I think, that's no big problem.
Could anyone review?

numberZero:
Panel on the right does work, while the one on the left doesn’t, despite receiving the same amount of daylight (plus some artificial light).
screenshot_20170430_001442

technic/machines/register/solar_array.lua
```diff
@@ -1,32 +1,28 @@
 
 local S = technic.getter
 
+local function get_light(param1)
+	local alight = (param1-param1%16)/16
+	local sunlight = (param1%16 ~= alight and param1%16) or 0
+	return sunlight, alight
```
HybridDog: artificial_light = math.floor(param1 / 16)
daylight = param1 % 16
sunlight : See https://github.com/HybridDog/technic/blob/hversion/technic/machines/register/solar_array_new.lua#L81

numberZero: This way, a solar panel in a closed box but under a mese lamp would work well. Actually that’s very similar to what technic currently does. I don't think it's time to change anything here until minetest/minetest#5155

sofar:
Why not use `math.max(sunlight, artificiallight)` ? that seems entirely logical to me....

HybridDog:
sofar, it doesn't work, see https://github.com/minetest/minetest/pull/5680#issuecomment-298316312

sofar:
This needs resolution - can we figure out a solution that everyone is happy with?

numberZero:
There is *no way* to separate sunlight and artificial light reliably (except of scanning the nodes around, as in  minetest/minetest#5680). See minetest/minetest#5155 for another suggestion. But either way, without engine change the PR’s title may not be done without making thin small solar panels big fat quarries, in terms of CPU and RAM usage.

numberZero:
Oh well, really... So, the Lua version of #5680 would be suitable for such usage; have you kept it?

HybridDog:
It's there: https://github.com/minetest/minetest/blob/468dd4d2cfaa8e2eaf012b0aa8f6d421b777f624/builtin/game/light_detection.lua

The4codeblocks:
Depends on the light source, emits UV? Generate power, and vice-versa

PR #360, opened 2017-04-25
Show progress in machines formspec

asakul:
default:furnace shows progress when processing something, so I figured I'd just implement it in the same way.

### DIFF
--- a/technic/machines/register/machine_base.lua
+++ b/technic/machines/register/machine_base.lua
@@ -65,18 +65,19 @@ function technic.register_base_machine(data)
 				"listring[current_name;upgrade2]"..
 				"listring[current_player;main]"
 		end
-		if tier ~= "LV" then
-			formspec_string = formspec_string..
-				fs_helpers.cycling_button(
-					meta,
-					pipeworks.button_base,
-					"splitstacks",
-					{
-						pipeworks.button_off,
-						pipeworks.button_on
-					}
-				)..pipeworks.button_label
+		if tier == "LV" then
+			return formspec_string
 		end
+		formspec_string = formspec_string..
+			fs_helpers.cycling_button(
+				meta,
+				pipeworks.button_base,
+				"splitstacks",
+				{
+					pipeworks.button_off,
+					pipeworks.button_on
+				}
+			)..pipeworks.button_label
 		return formspec_string
 	end
 
--- a/technic/machines/register/machine_base.lua
+++ b/technic/machines/register/machine_base.lua
@@ -43,8 +43,8 @@ function technic.register_base_machine(data)
 	local active_groups = {not_in_creative_inventory = 1}
 	for k, v in pairs(groups) do active_groups[k] = v end
 
-	local formspec = function(data, progress_percentage)
-		local formspec_string = "invsize[8,9;]"..
+	local function make_formspec(data, progress_percentage)
+		local formspec_string = "size[8,9;]"..
 			"list[current_name;src;"..(3.75-input_size)..",1.5;"..input_size..",1;]"..
 			"list[current_name;dst;5,1;2,2;]"..
 			"list[current_player;main;0,5;8,4;]"..
@@ -54,7 +54,7 @@ function technic.register_base_machine(data)
 			"listring[current_name;src]"..
 			"listring[current_player;main]"..
 			"image[3.75,1.5;1,1;gui_furnace_arrow_bg.png^[lowpart:"..
-			(progress_percentage)..":gui_furnace_arrow_fg.png^[transformR270]"
+			progress_percentage..":gui_furnace_arrow_fg.png^[transformR270]"
 		if data.upgrade then
 			formspec_string = formspec_string..
 				"list[current_name;upgrade1;1,3;1,1;]"..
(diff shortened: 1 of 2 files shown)


technic/machines/register/machine_base.lua
```diff
@@ -43,25 +43,29 @@ function technic.register_base_machine(data)
...
-		"listring[current_name;src]"..
-		"listring[current_player;main]"
-	if data.upgrade then
-		formspec = formspec..
-			"list[current_name;upgrade1;1,3;1,1;]"..
-			"list[current_name;upgrade2;2,3;1,1;]"..
-			"label[1,4;"..S("Upgrade Slots").."]"..
-			"listring[current_name;upgrade1]"..
+	local formspec = function(data, progress_percentage)
+		local formspec_string = "invsize[8,9;]"..
```
DS-Minetest: Why the indent here?

technic/machines/register/machine_base.lua
```diff
@@ -43,25 +43,29 @@ function technic.register_base_machine(data)
...
-		"listring[current_player;main]"..
-		"listring[current_name;src]"..
-		"listring[current_player;main]"
-	if data.upgrade then
-		formspec = formspec..
-			"list[current_name;upgrade1;1,3;1,1;]"..
-			"list[current_name;upgrade2;2,3;1,1;]"..
-			"label[1,4;"..S("Upgrade Slots").."]"..
-			"listring[current_name;upgrade1]"..
+	local formspec = function(data, progress_percentage)
```
DS-Minetest: local function formspec(data, progress_percentage)

technic/machines/register/machine_base.lua
DS-Minetest: `invsize` is deprecated, use `size`.

technic/machines/register/machine_base.lua
```diff
@@ -73,6 +77,14 @@ function technic.register_base_machine(data)
 		local machine_node      = "technic:"..ltier.."_"..machine_name
 		local machine_demand    = data.demand
 
+		local update_formspec = function(result)
```
DS-Minetest: Like above. `local function update_formspec(result)`

technic/machines/register/machine_base.lua
```diff
@@ -73,6 +77,14 @@ function technic.register_base_machine(data)
...
+			if result ~= nil then
```
DS-Minetest: Just `if result then` since it's a table.

technic/machines/register/machine_base.lua
```diff
@@ -43,25 +43,29 @@ function technic.register_base_machine(data)
...
+			"list[current_player;main;0,5;8,4;]"..
+			"label[0,0;"..machine_desc:format(tier).."]"..
+			"listring[current_name;dst]"..
 			"listring[current_player;main]"..
-			"listring[current_name;upgrade2]"..
-			"listring[current_player;main]"
+			"listring[current_name;src]"..
+			"listring[current_player;main]"..
+			"image[3.75,1.5;1,1;gui_furnace_arrow_bg.png^[lowpart:"..
+			(progress_percentage)..":gui_furnace_arrow_fg.png^[transformR270]"
```
DS-Minetest: Don't use the parenthesis(`(`...`)`).
(http://dev.minetest.net/Lua_code_style_guidelines#Misc 4th point)

DS-Minetest:
Actually this removes that new switch that pipeworks adds.
Also it doesn't make much sense for the furnaces since they are much faster than the arrow can be.
And could you also add this for the fuel fired alloy furnace?

asakul:
Yeah, it updates only once per second, furnaces just work too fast. However, I didn't find any way to make smooth animation in formspec.
Also, pipeworks switch is left as it is in fuel-fired furnace, and I can't see any switch in lv-furnace, even when using latest master (both pipeworks and technic).

DS-Minetest:
Yes, LV machines have no such switch since they are not connectable with tubes.

asakul:
Oh, you're right, it breaks them. Have to think of something ;)

technic/machines/register/machine_base.lua
```diff
@@ -43,28 +43,44 @@ function technic.register_base_machine(data)
...
+			formspec_string = formspec_string..
+				"list[current_name;upgrade1;1,3;1,1;]"..
+				"list[current_name;upgrade2;2,3;1,1;]"..
+				"label[1,4;"..S("Upgrade Slots").."]"..
+				"listring[current_name;upgrade1]"..
+				"listring[current_player;main]"..
+				"listring[current_name;upgrade2]"..
+				"listring[current_player;main]"
+		end
+		if tier ~= "LV" then
```
DS-Minetest:
```
if tier == "LV" then
	return formspec_string
end
.
.
.
```
Saves an indent.

numberZero:
Wouldn’t this make large (automated) systems (like my 34-centrifuge uranium enrichment facility) more laggy?
I think, that should be done with CSM. That would lag less, and could be much smoother that way.

asakul:
Yeah, looks like it scales pretty bad. Have to think of something again =)

Issue #363, opened 2017-04-29
corium_flowing has no side textures


DS-Minetest:
![screenshot_20170429_134146](https://cloud.githubusercontent.com/assets/7613443/25555281/34f91182-2ce2-11e7-99ed-8836c9f29331.png)
(Also as item it looks like an unknown node.)

numberZero:
Because its definition is too tricky. This is what MTG water uses:
```lua
        tiles = {"default_water.png"},
        special_tiles = {
                {
                        name = "default_water_flowing_animated.png",
                        backface_culling = false,
                        animation = {
                                type = "vertical_frames",
                                aspect_w = 16,
                                aspect_h = 16,
                                length = 0.8,
                        },
                },
                {
                        name = "default_water_flowing_animated.png",
                        backface_culling = true,
                        animation = {
                                type = "vertical_frames",
                                aspect_w = 16,
                                aspect_h = 16,
                                length = 0.8,
                        },
                },
        },
```
and this is corium:
```lua
                drawtype = (state == "source" and "liquid" or "flowingliquid"),
                [state == "source" and "tiles" or "special_tiles"] = {{
                        name = "technic_corium_"..state.."_animated.png",
                        animation = {
                                type = "vertical_frames",
                                aspect_w = 16,
                                aspect_h = 16,
                                length = 3.0,
                        },
                }},
```
So, see #364

DS-Minetest:
#364 is merged.

PR #364, opened 2017-04-29
Fix flowing corium textures

numberZero:
#363


numberZero:
Note: that trick with BF/non-BF tiles should be done by the engine IMO... But I just mimic what MTG does, as it works well.

PR #365, opened 2017-05-03
Actually reuse mapgen table

asl97:
See comments on https://github.com/minetest-mods/technic/commit/eb344ad804aa83570ccda9a5886733d8b42aa598#commitcomment-21588929

### DIFF
--- a/technic_worldgen/oregen.lua
+++ b/technic_worldgen/oregen.lua
@@ -118,7 +118,7 @@ minetest.register_ore({
 })
 
 -- Sulfur
-local sulfur_buf = nil
+local sulfur_buf = {}
 local sulfur_noise= nil
 
 minetest.register_on_generated(function(minp, maxp, seed)


HybridDog:
This should cause a crash. If the buffer is specified, the return value is nil. (`data` is set to it) [(according to that)](https://github.com/minetest-mods/technic/commit/eb344ad804aa83570ccda9a5886733d8b42aa598#commitcomment-21988845)
l guess you need to remove the data variable and use the buffer variable instead.

asl97:
@HybridDog did you even tried the patch?
It returned the buffer even if the buffer is specified but **[must be a table](https://github.com/minetest/minetest/blob/57e5aa662851485902575c3c747437e365bf72c8/src/script/lua_api/l_vmanip.cpp#L65)**.
ie: `data == sulfur_buf if sulfur_buf is table`

even @paramat said it in the post you linked: **"Looks like [they](https://github.com/minetest-mods/technic/commit/eb344ad804aa83570ccda9a5886733d8b42aa598#commitcomment-21988596) might be the same thing."**

Do you honestly think @numberZero and I would have suggested it without testing it first?
____
Here is a break down:
```c
	if (use_buffer)
		lua_pushvalue(L, 2); // this pushed the specified table into the stack
	else
		lua_newtable(L); // if no table is specified, it create a new table and push it into the stack

       // push stuff to the stack/table
	for (u32 i = 0; i != volume; i++) {
		lua_Integer cid = vm->m_data[i].getContent(); // get data from database
		lua_pushinteger(L, cid);  // push the data to the top of the stack
		lua_rawseti(L, -2, i + 1); // assign the top of the stack as value to the key `i+1` in the table at stack[-2],  ie: `stack[-2][i+1] = cid`
	}
       return 1; // return the top of the  stack, the same result for both with or without specifying the buffer
```

HybridDog:
> did you even tried the patch?

No, l assumed that [instead](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L3476) means in lieu of returning the table the passed buffer is used.

PR #368, opened 2017-05-13
Test 'have_ui' before calling 'unified_inventory' global

AntumDeluge:
Not sure if I did this right, but it fixes 'Undeclared global variable "unified_inventory"' warning.

### DIFF
--- a/technic/machines/register/recipes.lua
+++ b/technic/machines/register/recipes.lua
@@ -51,7 +51,7 @@ local function register_recipe(typename, data)
 	end
 	
 	technic.recipes[typename].recipes[index] = recipe
-	if unified_inventory and technic.recipes[typename].output_size == 1 then
+	if have_ui and technic.recipes[typename].output_size == 1 then
 		unified_inventory.register_craft({
 			type = typename,
 			output = data.output,


DS-Minetest:
Nope, this is really not needed. Such nests are never needed in lua. The things right next to `and` are ignored if the thing left is nil or false.

Removing this warning makes sense.
The only thing, you've really changed is that you have added the `have_ui` down there. Just include that with an `and` and rename the topic title.

AntumDeluge:
How is that? Or should I overwrite branch so previous commit isn't in history?

technic/machines/register/recipes.lua
```diff
@@ -51,7 +51,7 @@ local function register_recipe(typename, data)
 	end
 	
 	technic.recipes[typename].recipes[index] = recipe
-	if unified_inventory and technic.recipes[typename].output_size == 1 then
+	if have_ui and unified_inventory.register_craft_type and data.output_size == 1 then
```
DS-Minetest: I think, just changing `unified_inventory` to `have_ui` would be enough.
Btw. `have_ui` could be set instead of path to `minetest.global_exists("unified_inventory")`.

DS-Minetest:
Could this trivial fix please be merged sometime?

Issue #372, opened 2017-06-23
Placing a chest with items crashes


VanessaE:
```
2017-06-23 20:29:05: ERROR[Main]: ServerError: AsyncErr: ServerThread::run Lua: Runtime error from mod '' in callback item_OnPlace(): ...val_World/worldmods/dreambuilder_modpack/wrench/init.lua:45: attempt to index local 'data' (a nil value)
2017-06-23 20:29:05: ERROR[Main]: stack traceback:
2017-06-23 20:29:05: ERROR[Main]:       ...val_World/worldmods/dreambuilder_modpack/wrench/init.lua:45: in function 'after_place_node'
2017-06-23 20:29:05: ERROR[Main]:       /usr/share/minetest/builtin/game/item.lua:307: in function </usr/share/minetest/builtin/game/item.lua:208>
```

Before it actually goes down, it manages to place a chest that has no inventory defined at all (e.g. nil), no meta, and node properties that prevent even a maptools admin pick from digging it.  In this case, it was a picked-up silver chest containing some items.

DS-Minetest:
When was the chest picked up?
I can't reproduce it.

VanessaE:
Unfortunately I have no info on that.  All I know is that it was a normal user (i.e. no special privs), and that placing the picked-up chest still leaves a copy in the player's inventory.  The chest also caused a crash in Pipeworks' filter-injector, due to having nil inventory, but I patched around that.

DS-Minetest:
Hm, :thinking:, how could `data` not be a serialized table¿
Maybe it would be good not to place the chest if `data` is nil.

VanessaE:
Perhaps, but the user did say they were sure there was at least some marble in it.

AntumDeluge:
Just to confirm, I am having the same issue(s) (I think):

```
2017-06-28 17:31:01: ERROR[Main]: ServerError: AsyncErr: ServerThread::run Lua: Runtime error from mod '' in callback item_OnPlace(): ...ames/antum-testing/mods/modpacks/technic/wrench/init.lua:45: attempt to index local 'data' (a nil value)
2017-06-28 17:31:01: ERROR[Main]: stack traceback:
2017-06-28 17:31:01: ERROR[Main]: 	...ames/antum-testing/mods/modpacks/technic/wrench/init.lua:45: in function 'after_place_node'
[...]

PR #373, opened 2017-06-23
don't place eg. chests without data

DS-Minetest:
Related: #372


VanessaE:
I can at least confirm this stops the crash.  I would still like to know why some chests have no "data" in the first place though.

DS-Minetest:
https://github.com/minetest-mods/technic/commit/16146231b8f8c8562c1d1fd4a2c84bcf4134c3c5

PR #382, opened 2017-07-08
Replace deprecated 'intllib' function:

AntumDeluge:
- 'intllib.Getter' with 'intllib'make_gettext_pair' if available,
otherwise use deprecated function.

### DIFF
--- a/concrete/init.lua
+++ b/concrete/init.lua
@@ -5,7 +5,7 @@ local technic = rawget(_G, "technic") or {}
 technic.concrete_posts = {}
 
 -- Boilerplate to support localized strings if intllib mod is installed.
-local S = rawget(_G, "intllib") and intllib.Getter() or function(s) return s end
+local S = rawget(_G, "intllib") and intllib.make_gettext_pair() or function(s) return s end
 
 for i = 0, 31 do
 	minetest.register_alias("technic:concrete_post"..i,
--- a/technic/machines/MV/lighting.lua
+++ b/technic/machines/MV/lighting.lua
@@ -7,7 +7,7 @@
 local S
 if (minetest.get_modpath("intllib")) then
 	dofile(minetest.get_modpath("intllib").."/intllib.lua")
-	S = intllib.Getter(minetest.get_current_modname())
+	S = intllib.make_gettext_pair(minetest.get_current_modname())
 else
 	S = function (s) return s end
 end
--- a/extranodes/init.lua
+++ b/extranodes/init.lua
@@ -1,7 +1,7 @@
 -- Minetest 0.4.6 mod: extranodes
 -- namespace: technic
 -- Boilerplate to support localized strings if intllib mod is installed.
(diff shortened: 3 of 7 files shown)


concrete/init.lua
```diff
@@ -5,7 +5,7 @@ local technic = rawget(_G, "technic") or {}
 technic.concrete_posts = {}
 
 -- Boilerplate to support localized strings if intllib mod is installed.
-local S = rawget(_G, "intllib") and intllib.Getter() or function(s) return s end
+local S = rawget(_G, "intllib") and intllib.make_gettext_pair() or intllib.Getter() or function(s) return s end
```
numberZero: `rawget(_G, "intllib") and (intllib.make_gettext_pair() or intllib.Getter()) or function(s) return s end`

AntumDeluge: Oh jeez, I can't believe I forgot the braces. :roll_eyes:

AntumDeluge:
How is that?

AntumDeluge:
Would it be better to replace ```technic.getter = intllib.Getter()``` with ```technic.getter = (intllib.make_gettext_pair() or intllib.Getter())``` instead of using the condition checks?

numberZero:
> Would it be better to replace technic.getter = intllib.Getter() with technic.getter = (intllib.make_gettext_pair() or intllib.Getter()) instead of using the condition checks?

@AntumDeluge no, conditional statements are better. What I suggested is wrong: braces per se can’t help if `make_gettext_pair` is not defined, and checking its existence as well makes the statement insanely long.

concrete/init.lua
```diff
@@ -5,7 +5,7 @@ local technic = rawget(_G, "technic") or {}
 technic.concrete_posts = {}
 
 -- Boilerplate to support localized strings if intllib mod is installed.
-local S = rawget(_G, "intllib") and intllib.Getter() or function(s) return s end
+local S = rawget(_G, "intllib") and (intllib.make_gettext_pair() or intllib.Getter()) or function(s) return s end
```
SmallJoker: This does not do what it should (see numberZero's comment). Just use `intllib.make_gettext_pair()` everywhere (exists for an entire year already). Do not support legacy code and let them update their outdated mods.

technic/machines/MV/lighting.lua
```diff
@@ -7,7 +7,11 @@
 local S
 if (minetest.get_modpath("intllib")) then
 	dofile(minetest.get_modpath("intllib").."/intllib.lua")
-	S = intllib.Getter(minetest.get_current_modname())
+	if (intllib.make_gettext_pair) then
```
SmallJoker: `S = technic.getter` or am I missing something?

AntumDeluge:
It's been a while. I hope that is what you were looking for. Let me know if I misunderstood anything please.

SmallJoker:
Closing in favour of #529

Issue #355, opened 2017-04-15
power supply balancing


DS-Minetest:
IMHO infinite power sources (eg. watermill, solar array) are too powerful relative to not infinite ones (like fuel fired generator or nuclear reactor).
Also solar arrays and panels are too strongly affected by place hight (pos.y).
I wouldn't make the non infinite power sources stronger but the infinite ones weaker since technic power seems to be a bit cheap.

kaeza:
[Here](http://irc.minetest.net/minetest/2017-04-06#i_4859574) is some earlier conversation regarding this.

DS-Minetest:
We could give the generators an inventory where you have to put in a blade (or something else), that gets more wear over time.

Thomas--S:
I'm fine with making the infinite ones a little bit weaker, but I'd like to still get an acceptable amount of power from them. (I think half of the current power may be ok, anything that goes beyond this wouldn't be so good IMHO)

DS-Minetest:
IMHO 1/4 would be better.
So, eg.:

name|now|1/2|1/4
-|-|-|-
LV solar panel|200|100|50
LV geothermal|300|150|75
LV fuel-fired|200||
HV nuclear reactor|10000

Hm I think, less reduction would be ok. But **solar has to be much more limited in altitude**.

Thomas--S:
I have no problem with the solar power change. I'd only like to see that the hydro generators after the change still provide plenty of power.

numberZero:
@Thomas--S But I’d like to see different hydro generators: small, cheap (no diamonds! only copper!), low-power (10–100 PU) LV generator, and large, expensive, powerful (1–10 kPU) MV and/or HV generator.
@DS-Minetest You seem to miss a zero: HV nuclear reactor 	power is 100000 (100 kPU) =)

Thomas--S:
That's a really good idea. :+1:

DS-Minetest:
Water mill gives a maximum supply of 1575.

DustyBagel:
I would like to see solar panels power output more affected by cloudy skies caused by other mods like snowdrift. Also, I think it doesn't make sense that 3 technic:solar_panel nodes make more power than 1 arrayed lv solar panel because it takes 3 lv solar panels to make 1 arrayed lv solar panel and you save on the extra resources it takes to make the arrayed lv panel.

[...]

PR #383, opened 2017-07-25
rebalance a bit

Desour:
Related: #355 
`changes.txt` has to be removed.
- solar:
solar_panel_supply
(y-axis is hight, x-axis is power supply, light level is constantly at maximum)
- Water mills have again a maximum of 200 and gain 25% for each side with water with param2 = 15.
- LV grinder and furnaces need twice as much power.
- Fuel-fired generators supply twice as much power.
- No more big changes.


SmallJoker:
What are the opinions of other players about these changes? Not to forget that the craft recipes aren't too cheap to set up a circuit.

PR #388, opened 2017-08-29
nuclear_reactor: do not set structure_accumulated_badness to less than 0

### LINKED ISSUE #350: old nuclear reactors last longer than new ones if hull breaks
The problem is somewhere [here](https://github.com/minetest-mods/technic/blob/a84b5f3b28789318738dfa52ee1264c8d31eecb4/technic/machines/HV/nuclear_reactor.lua#L259).
The meta int "structure_accumulated_badness" becomes negative. (eg. -4268)
And so, there's a very big buffer until it reaches 25.

DS-Minetest:
Fixes #350.


sofar:
:+1: lgtm

DS-Minetest:
To be merged? This has one legitimisation and two yellow approvals.

sofar:
Give it a week more and ping me if it's not merged by then.

PR #390, opened 2017-09-05
some craft fixes

DS-Minetest:
- The new sandstone types (desert and silver) are added to grinder and compressor craft.
- Some crafting was cleared. Note: unified_inventory may show it wrong.

### DIFF
--- a/technic/machines/register/compressor_recipes.lua
+++ b/technic/machines/register/compressor_recipes.lua
@@ -11,6 +11,8 @@ end
 local recipes = {
 	{"default:snowblock",          "default:ice"},
 	{"default:sand 2",             "default:sandstone"},
+	{"default:desert_sand 2",      "default:desert_sandstone"},
+	{"default:silver_sand 2",      "default:silver_sandstone"},
 	{"default:desert_sand",        "default:desert_stone"},
 	{"technic:mixed_metal_ingot",  "technic:composite_plate"},
 	{"default:copper_ingot 5",     "technic:copper_plate"},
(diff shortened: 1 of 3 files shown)


technic/machines/register/compressor_recipes.lua
```diff
@@ -21,10 +23,21 @@ local recipes = {
 
 -- defuse the default sandstone recipe, since we have the compressor to take over in a more realistic manner
 minetest.clear_craft({
-	output = "default:sandstone",
 	recipe = {
-		{'group:sand', 'group:sand'},
-		{'group:sand', 'group:sand'}
```
numberZero: shouldn’t the group recipe be cleared as well?

technic/machines/register/extractor_recipes.lua
```diff
@@ -36,20 +36,32 @@ if minetest.get_modpath("dye") then
 	-- overwrite the existing crafting recipes
 	local dyes = {"white", "red", "yellow", "blue", "violet", "orange"}
 	for _, color in ipairs(dyes) do
+		minetest.clear_craft({
+			type = "shapeless",
+			recipe = {"group:flower,color_"..color},
+		})
```
numberZero: Is this here to make `unified_inventory` show correct recipes?

technic/machines/register/compressor_recipes.lua
DS-Minetest: The group recipe is old and would give an error. Look [here](https://github.com/minetest/minetest_game/blob/master/mods/default/crafting.lua#L465L471).

technic/machines/register/extractor_recipes.lua
DS-Minetest: No, `unified_inventory` still shows the recipes. But they are removed because there might be ways to access the old recipe without this, eg. with a [crafting table](https://forum.minetest.net/viewtopic.php?t=15405).

technic/machines/register/compressor_recipes.lua
numberZero: Or [here](https://github.com/minetest/minetest/blob/master/src/script/lua_api/l_craft.cpp#L350L351)... I though it’s not an error to clear a non-existent recipe, sorry.

technic/machines/register/extractor_recipes.lua
SmallJoker: Late review - but why do you need to clear the craft recipe? Is it to ensure the correct dye recipe output which is different to `"dye:"..color`?

t4im: It's to justify the existence of the extractor as game play mechanic. Back then the extractor produced only half the output of the default-dye crafting recipes and you were better off just using the autocrafter. I boosted the extractor output and left a low crafting output to encourage extractor use without risking unresolvable recipe dependencies in the future.

PR #391, opened 2017-09-06
Replace old nodeupdate() with new minetest.check_for_falling()

Thomas--S:
Otherwise technic crashes on 0.5.0-dev.

This is not tested, but should work.


beyondlimits:
Bump!

I hit that issue too!

Issue #392, opened 2017-09-10
Wrench: allow players with protection_bypass to use wrench on other player stuff


Fixer-007:
This can be used by admins to move other people locked chests/other without hassle. Upvote if you support this.

SmallJoker:
```patch
From d2d88692e363e3896c2a22d7482348f28224789b Mon Sep 17 00:00:00 2001
From: SmallJoker <mk939@ymail.com>
Date: Fri, 22 Sep 2017 16:56:00 +0200
Subject: [PATCH] Wrench: Allow picking up with protection_bypass privilege

---
 wrench/init.lua | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/wrench/init.lua b/wrench/init.lua
index f99fd14..a45dd1f 100644
--- a/wrench/init.lua
+++ b/wrench/init.lua
@@ -115,7 +115,7 @@ minetest.register_tool("wrench:wrench", {
 			return
 		end
 		local meta = minetest.get_meta(pos)
-		if def.owned then
+		if def.owned and not minetest.check_player_privs(player, "protection_bypass") then
 			local owner = meta:get_string("owner")
 			if owner and owner ~= player_name then
 				minetest.log("action", player_name..
-- 
2.7.4
```

Fixer-007:
Works for me!

Fixer-007:
Made a PR here: https://github.com/minetest-mods/technic/pull/393

PR #393, opened 2017-09-22
Wrench: Allow picking up with protection_bypass privilege

Fixer-007:
This allows users with protection_bypass to move other players locked chests/etc without hassle (relocating flying chests).  Currently it is not possible, you can move only your own chests.


DS-Minetest:
:+1: 
You could add "Fixes #392." to the first comment.

Fixer-007:
Added

Fixer-007:
Almost 1 year later, but thanks! :D

PR #395, opened 2017-10-26
Lawn Trimmer (aka Cord Trimmer or Weed Whacker) tool

h-v-smacker:
This adds a common gardening tool to the technic mod. This tool, apart from its direct purpose (which it serves very well, even though it has mostly aesthetic meaning in minetest), serves two quite practical goals: to aid in finding plants that can be cultured in the wilderness and to help collect flowers when producing organic dye pigments. It has four modes of operation, from mowing directly under the operator's feet, to sweeping 3 blocks around.

The rationale is to provide an auxiliary tool for dealing with plants, much like the vacuum tool comes in handy when chopping down trees (or cutting them with the chainsaw) and collecting the falling blocks, fruits, and saplings.


technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+	It comes with 4 modes of operation, defined by how wide its sweep is: from 0 (at one's feet)
+	to 3 nodes in radius (square radius, as most things in minetest are).
+
+	The sound is an edited fragment from https://www.cutestockfootage.com/sound-effect/9251/grass-trimmer-01
+	used in accordance with its licensing terms (free use for any purpose in an altered form)
+]]
+
+-- Configuration
+local lawn_trimmer_max_charge        = 10000 -- 10000 - Maximum charge of the lawn trimmer
+local lawn_trimmer_charge_per_object = 25    -- 25    - Can mow 400 'group:flora' blocks
```
numberZero: Text style:
1. Tautology
2. -- Tautology

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+	{S("immediately around the user")},
+	{S("sweep 1 block wide")},
+	{S("sweep 2 blocks wide")},
+	{S("sweep 3 blocks wide")}
+}
+
+
+-- Mode switcher for the tool
+local function lawn_trimmer_setmode(user,itemstack)
+	local player_name=user:get_player_name()
```
numberZero: Spaces, spaces! (everywhere in this function)

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+	local item=itemstack:to_table()
+	local meta=minetest.deserialize(item["metadata"])
```
numberZero: `itemstack:get_metadata()`

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+	end
+	if meta["mode"]==nil then
+		minetest.chat_send_player(player_name, S("Use while sneaking to change Lawn Trimmer modes."))
+		meta["mode"]=0
+		mode=0
+	end
+	mode=(meta["mode"])
+	mode=mode+1
+	
+	if mode>4 then mode=1 end
```
numberZero: Er, what? Rethink this part (from `local mode`)

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+	itemstack:set_metadata(minetest.serialize(meta))
+	return itemstack
+end
+
+
+-- Perform the trimming action
+local function trim_the_lawn(itemstack, user)
+	local meta = minetest.deserialize(itemstack:get_metadata())
+	local keys = user:get_player_control()
+	local meta = minetest.deserialize(itemstack:get_metadata())
```
numberZero: Second `local meta`

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+	end
+	
+	local pos = user:getpos()
+	local inv = user:get_inventory()
+	-- Defining the area for the search needs two positions
+	local start_pos = user:getpos()
+	start_pos.x = start_pos.x - meta.mode + 1
+	start_pos.z = start_pos.z - meta.mode + 1
+	start_pos.y = start_pos.y - 1 -- the tool can hardly be lowered a lot while standing, but we'll allow 1 node
+	local end_pos = user:getpos()
```
numberZero: 3rd `getpos` in a row.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,156 @@
...
+local function lawn_trimmer_setmode(user,itemstack)
+	local player_name=user:get_player_name()
+	local item=itemstack:to_table()
+	local meta=minetest.deserialize(item["metadata"])
+	local mode
+	if meta==nil then
+		meta={}
+		mode=0
+	end
+	if meta["mode"]==nil then
```
numberZero: `meta.mode`

technic/tools/lawn_trimmer.lua
h-v-smacker: The commented out sections represent the suggested defaults (the same is done in vacuum.lua)

technic/tools/lawn_trimmer.lua
h-v-smacker: I beg your pardon?

technic/tools/lawn_trimmer.lua
h-v-smacker: Indeed, it can be done like that, but this is exactly how it's done in all other tools in technic. I thought there was a reason for such code style.

technic/tools/lawn_trimmer.lua
h-v-smacker: The only tool with mode switching is the mining drill, and that's how it's done there.

technic/tools/lawn_trimmer.lua
h-v-smacker: Again, the same criticism can be addressed to the mining drill implementation.

technic/tools/lawn_trimmer.lua
h-v-smacker: Oh, that's a good one, my bad.

technic/tools/lawn_trimmer.lua
h-v-smacker: Also a good one.

technic/tools/lawn_trimmer.lua
benrob0329: http://dev.minetest.net/Lua_code_style_guidelines

h-v-smacker: Much appreciated, will review my code again to ensure compliance.

numberZero: There should be spaces around `=` (and most other operators).

technic/tools/lawn_trimmer.lua
```diff
@@ -20,42 +20,39 @@
 ]]
 
 -- Configuration
-local lawn_trimmer_max_charge        = 10000 -- 10000 - Maximum charge of the lawn trimmer
-local lawn_trimmer_charge_per_object = 25    -- 25    - Can mow 400 'group:flora' blocks
+local lawn_trimmer_max_charge        = 10000 -- default: 10000 - Same as the chainsaw
+local lawn_trimmer_charge_per_object = 25    -- default: 25    - Can mow 400 'group:flora' blocks
```
numberZero:
```lua
local lawn_trimmer_max_charge = 10000

-- Can mow 400 'group:flora' blocks on a single charge
local lawn_trimmer_charge_per_object = 25
```

technic/tools/lawn_trimmer.lua
numberZero: > the same is done in vacuum.lua

Not that good reason to do the same, considering the overall code quality around. That’s a bad practice to force to edit the code to change settings. You may notice that even in `technic` it’s not used everywhere, there is a setting system.

technic/tools/lawn_trimmer.lua
numberZero: So, don’t copy-paste blindly :P

h-v-smacker: Lesson learned. I reviewed the code to comply with the guidelines. Hopefully, I haven't missed much this time.

technic/tools/lawn_trimmer.lua
h-v-smacker: I believe this part can be left as it is for now though, because such changes as centralized configuration should be introduced to all the tools at once.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+		description = S("Lawn Trimmer Mode %d"):format(i),
+		inventory_image = "technic_lawn_trimmer.png^technic_tool_mode" .. i .. ".png",
+		wield_image = "technic_lawn_trimmer.png",
+		wear_represents = "technic_RE_charge",
+		on_refill = technic.refill_RE_charge,
+		groups = {not_in_creative_inventory = 1},
+		on_use = function(itemstack, user, pointed_thing)
+			trim_the_lawn(itemstack, user)
+			return itemstack
+		end
```
DS-Minetest: `on_use = trim_the_lawn(itemstack, user),`

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	})
+end
+
+
+-- Provide a crafting recipe
+local mesecons_button = minetest.get_modpath("mesecons_button")
+local trigger = mesecons_button and "mesecons_button:button_off" 
+	or "default:mese_crystal_fragment"
```
DS-Minetest: There's no need for a `mesecons_button` variable:
```lua
local trigger = minetest.get_modpath("mesecons_button") and
	"mesecons_button:button_off" or "default:mese_crystal_fragment"
```

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	stack_max = 1,
+	wear_represents = "technic_RE_charge",
+	on_refill = technic.refill_RE_charge,
+	on_use = function(itemstack, user, pointed_thing)
+		trim_the_lawn(itemstack, user)
+		return itemstack
+	end
+})
+
+for i = 1,4,1 do
```
DS-Minetest: `for i = 1, 4 do`

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+local lawn_trimmer_mode_text = {
+	S("immediately around the user"),
+	S("sweep 1 block wide"),
+	S("sweep 2 blocks wide"),
+	S("sweep 3 blocks wide")
+}
+
+
+-- Mode switcher for the tool
+local function lawn_trimmer_setmode(user,itemstack)
```
DS-Minetest: Missing space after `,`.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	local player_name = user:get_player_name()
+	local meta = minetest.deserialize(itemstack:get_metadata())
```
DS-Minetest: Metadata is deprecated, use itemstack meta instead: https://github.com/minetest/minetest/blob/610ea6f21629933be3b38c2b5fce53a35c5aac3a/doc/lua_api.txt#L3690

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+		meta = {mode = 0}
+	end
+	if not meta.mode then
+		minetest.chat_send_player(player_name, 
+			S("Use while sneaking to change Lawn Trimmer modes."))
+		meta.mode = 0
+	end
+	
+	meta.mode = meta.mode + 1
+	if meta.mode > 4 then meta.mode = 1 end
```
DS-Minetest: `mode = mode % 4 + 1`

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	end
+	
+	if meta.charge > lawn_trimmer_charge_per_object then
+		minetest.sound_play("technic_lawn_trimmer", {
+			to_player = user:get_player_name(),
+			gain = 0.4,
+		})
+	end
+	
+	local pos = user:getpos()
```
DS-Minetest: `get_pos()`

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	end
+	if not meta or not meta.charge then
+		return
+	end
+	
+	if meta.charge > lawn_trimmer_charge_per_object then
+		minetest.sound_play("technic_lawn_trimmer", {
+			to_player = user:get_player_name(),
+			gain = 0.4,
+		})
```
DS-Minetest: Imo other players around should hear this, too.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	return itemstack
+end
+
+
+-- Perform the trimming action
+local function trim_the_lawn(itemstack, user)
+	local meta = minetest.deserialize(itemstack:get_metadata())
+	local keys = user:get_player_control()
+	
+	if not meta or not meta.mode or keys.sneak then
```
DS-Minetest: Maybe move `user:get_player_control()` down here, keys is only used once.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+		return lawn_trimmer_setmode(user, itemstack)
+	end
+	if not meta or not meta.charge then
+		return
+	end
+	
+	if meta.charge > lawn_trimmer_charge_per_object then
```
DS-Minetest: You should abort here if `meta.charge < lawn_trimmer_charge_per_object`.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+	local found_flora = minetest.find_nodes_in_area(start_pos, end_pos, {"group:flora"})
+	for _,f in ipairs(found_flora) do
+		-- Abort if no charge left for a node
+		if meta.charge < lawn_trimmer_charge_per_object then break end
+		-- Skip to the next one if this node cannot be dug
+		if not minetest.is_protected(f, user:get_player_name()) then
+			meta.charge = meta.charge - lawn_trimmer_charge_per_object
+			local node = minetest.get_node_or_nil(f)
+			minetest.node_dig(f, node, user)
+		end
```
DS-Minetest: Afaik `node_dig` already checks if the user can dig the node.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+		z = pos.z + meta.mode - 1,
+		y = pos.y + 1
+	} 
+
+	-- Since nodes sometimes cannot be removed, we cannot rely on repeating 
+	-- find_node_near() and removing found nodes
+	local found_flora = minetest.find_nodes_in_area(start_pos, end_pos, {"group:flora"})
+	for _,f in ipairs(found_flora) do
+		-- Abort if no charge left for a node
+		if meta.charge < lawn_trimmer_charge_per_object then break end
```
DS-Minetest: Move this beyond `node_dig`.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,169 @@
...
+minetest.register_tool("technic:lawn_trimmer", {
+	description = S("Lawn Trimmer"),
+	inventory_image = "technic_lawn_trimmer.png",
+	stack_max = 1,
+	wear_represents = "technic_RE_charge",
+	on_refill = technic.refill_RE_charge,
+	on_use = function(itemstack, user, pointed_thing)
+		trim_the_lawn(itemstack, user)
+		return itemstack
+	end
```
DS-Minetest: `on_use = trim_the_lawn,`

technic/tools/lawn_trimmer.lua
h-v-smacker: All other tools use deserialization. In fact, if I switch to set_int/get_int, I run into issues. This is how the tool's metadata look like after it is charged (so whatever the charge is, it's set by code which is already in technic mode) - `get_meta()` → `to_table()`:

`{["fields"] = {[""] = "return {[\"charge\"] = 10000}"}}`

and this is what happens when I get_meta() of an itemstack and then set_int("mode") of it:

```
local meta = itemstack:get_meta()
meta:set_int("mode", 1)
```
`{["fields"] = {[""] = "return {[\"charge\"] = 10000}", ["mode"] = "1"}}`

So charge gets into a serialized string stored in an unnamed field, and to access it through `get_meta()`, both converting to table and deserialization would be required all the same. This, in turn, will make it a literal mess, which would just make it harder to bring it into orderly condition.

technic/tools/lawn_trimmer.lua
h-v-smacker: (a) Some people find the sound annoying, and (b) the tool presents no danger or concern to other players, so making them listen to another random sound is literally pointless, it bears no useful information.

technic/tools/lawn_trimmer.lua
h-v-smacker: I prefer to group local declarations.

technic/tools/lawn_trimmer.lua
h-v-smacker: I cannot find definitive proof for that. In fact, the page http://dev.minetest.net/minetest.node_dig illustrates overriding the function to add checks for area protection, and http://dev.minetest.net/minetest.register_node#on_dig says node_dig only checks node.diggable, which is the node's own property independent of protection status.

technic/tools/lawn_trimmer.lua
DS-Minetest: I forgot, that battery boxes don't load in itemstack meta charge but in metadata charge. >_<
But that should be changed. (https://github.com/minetest-mods/technic/blob/e056800cea806be2a16476e338d8f8df9d8520ba/technic/machines/register/battery_box.lua#L425)
Ok, as long as there's no change in technic's api, this code is ok…

technic/tools/lawn_trimmer.lua
DS-Minetest: (a) Annoying sounds are funny. (b) Playing a sound is not always about giving useful information.

technic/tools/lawn_trimmer.lua
h-v-smacker: Changes are, indeed, necessary. But having all tools in one style would make subsequent changes easier, because one wouldn't have to figure out some "interim hacks".

However, the tools should have their code cleaned up first as well. Most of the criticisms I received were related to the code I adapted from the other existing tools, believing I could use them as a template of quality which is acceptable to the project. Turns out, that code is subpar.

technic/tools/lawn_trimmer.lua
DS-Minetest: The dev wiki isn't good. Better look to `lua_api.txt`.
And source code always tells the truth: https://github.com/minetest/minetest/blob/a5d5728f241fec8644ba65cb3b98c7dd1be2eee3/builtin/game/item.lua#L548L555

technic/tools/lawn_trimmer.lua
h-v-smacker: I combined the initial charge check with playing the sound, it seemed most fitting.

DS-Minetest: Meh, you can play the sound after the `end`, it would save 3 indents and one line.

technic/tools/lawn_trimmer.lua
h-v-smacker: How do I check if the node was actually dug though? Otherwise, I have no idea how many charge to spend. Currently, I use a reasonable assumption that, barring protection violations, there is no reason for a member of group flora to be undiggable (to the best of my knowledge, there are no steel flowers). Meanwhile, `node_dig` returns `nil` either way, regardless of whether the node was dug or not.

DS-Minetest: Use the `after_use` callback.

technic/tools/lawn_trimmer.lua
DS-Minetest: There's much old code in technic.

technic/tools/lawn_trimmer.lua
h-v-smacker: So instead of doing one redundant check, I should do a callback each time a node is dug, replacing the original itemstack with the returned one from the callback? What's the gain here?

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,171 @@
...
+local node_removed
+
+-- Mode switcher for the tool
+local function lawn_trimmer_setmode(user, itemstack)
+	local player_name = user:get_player_name()
+	local meta = minetest.deserialize(itemstack:get_metadata())
+
+	if not meta then
+		meta = {mode = 0}
+	end
```
DS-Minetest: `meta = meta or {mode = 0}` would be shorter.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,171 @@
...
+	S("sweep 2 blocks around the user"),
+	S("sweep 3 blocks around the user")
+}
+
+local node_removed
+
+-- Mode switcher for the tool
+local function lawn_trimmer_setmode(user, itemstack)
+	local player_name = user:get_player_name()
+	local meta = minetest.deserialize(itemstack:get_metadata())
```
DS-Minetest: You are getting `meta` twice, give `meta` instead as function parameter.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,171 @@
...
+
+-- Perform the trimming action
+local function trim_the_lawn(itemstack, user)
+	local meta = minetest.deserialize(itemstack:get_metadata())
+	local keys = user:get_player_control()
+	
+	if not meta or not meta.mode or keys.sneak then
+		return lawn_trimmer_setmode(user, itemstack)
+	end
+	if not meta or not meta.charge then
```
DS-Minetest: `not meta` will never happen here.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,171 @@
...
+	if not technic.creative_mode then
+		technic.set_RE_wear(itemstack, meta.charge, lawn_trimmer_max_charge)
+		itemstack:set_metadata(minetest.serialize(meta))
+	end
+	return itemstack
+end
+
+function check_removal(itemstack, user, node, digparams)
+	node_removed = true
+	return itemstack
```
DS-Minetest: (You could also return `nil` or `false`.)

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,171 @@
...
+technic.register_power_tool("technic:lawn_trimmer", lawn_trimmer_max_charge)
+minetest.register_tool("technic:lawn_trimmer", {
+	description = S("Lawn Trimmer"),
+	inventory_image = "technic_lawn_trimmer.png",
+	stack_max = 1,
+	wear_represents = "technic_RE_charge",
+	on_refill = technic.refill_RE_charge,
+	on_use = trim_the_lawn,
+	after_use = check_removal
+
```
DS-Minetest: Empty line.

technic/tools/lawn_trimmer.lua
h-v-smacker: Is this Perl golf or something?

technic/tools/lawn_trimmer.lua
h-v-smacker: Why would anyone return an expression that evaluates to false on success?

technic/tools/lawn_trimmer.lua
DS-Minetest: What is Perl golf?

technic/tools/lawn_trimmer.lua
DS-Minetest: The source code _says_, you don't have to return an itemstack. It does `wielded = after_use(…) or wielded`, ergo you can return a false value (`nil` or `false`) and it won't replace the itemstack. And `return nil` would be the same as `return` or not returning, ergo you can save one line.
But because this is not important, I used the braces.
(Btw you can also remove the parameters.)

technic/tools/lawn_trimmer.lua
h-v-smacker: It's when Perl programmers compete to write the shortest program to do a given task.

DS-Minetest: Ok, good to know.
This is no Perl golf since lua is used and there's no competition.
I just like compactness.

technic/tools/lawn_trimmer.lua
```diff
@@ -0,0 +1,168 @@
...
+-- Mode switcher for the tool
+local function lawn_trimmer_setmode(user, itemstack, meta)
+	local player_name = user:get_player_name()
+
+	if not meta then
+		meta = {mode = 0}
+	end
+	if not meta.mode then
+		minetest.chat_send_player(player_name,
```
numberZero: Won’t be called, as 0 is evaluated to **true** in Lua.
Maybe merge these `if` clauses?

h-v-smacker: On the other hand, something else can hypothetically change the object's metadata between placing it in the inventory and using it for the first time. In that case its metadata will have some fields, but not "mode". Better initiate empty meta with `mode = nil` then.

numberZero: > charging box

sorry, forgot that...

> `meta = {mode = nil}`

But `nil` means “no value” in Lua, thus that’s the same as `meta = {}`.
BTW, `meta = meta or {}` is acceptable IMO, it’s a conventional way to specify a default.

numberZero:
By the way, is there anyone around who could merge this?

PR #401, opened 2017-11-28
Use tin from MTG instead of `moreores`

numberZero:
minetest/minetest_game@26686196383df71e5634ee4c5af479668d83ee7e has tin so that should be used instead of `moreores`’ tin.


DS-Minetest:
:+1: 
But you broke some space alignments.

numberZero:
@DS-Minetest Sorry, I use variable-width font. And I consider space alignment evil.
Nevertheless, should I fix it?

SmallJoker:
@numberZero Yes, it would be nice to have them all indented in the same column.

PR #406, opened 2018-01-18
(minor) Grammar nazi corrections and straightening of facts within manual.md

ChimneySwift:
- There are manuals for all of the modpacks Technic depends on (possibly this is recent)
- Tin is now part of minetest_game
- Upon reading through I found some minor typos

### DIFF
--- a/manual.md
+++ b/manual.md
@@ -14,8 +14,13 @@ The technic modpack depends on some other modpacks:
 *   pipeworks, which supports the automation of item transport
 *   moreores, which provides some additional ore types
 
-This manual doesn't explain how to use these other modpacks, which ought
-to (but actually don't) have their own manuals.
+This manual doesn't explain how to use these other modpacks, which have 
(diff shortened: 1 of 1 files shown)


manual.md
```diff
@@ -68,10 +73,11 @@ own for its electrical conductivity, or as the base component of alloys.
 Although common, it is very heavily used, and most of the time it will
 be the material that most limits your activity.
 
-Tin is supplied by the moreores mod.  It is found from elevation +8
-downwards, with no elevation-dependent variations in abundance beyond
-that point.  It is a common metal.  Its main use in pure form is as a
-component of electrical batteries.  Apart from that its main purpose is
+Tin is part of the basic Minetest game (having migrated there from
```
SmallJoker: This is not true (yet). Might be best to remove this sentence except for the case that #401 is merged before this PR (would delay it).

PR #407, opened 2018-01-27
Remove unused values

numberZero:
The VManip was not used since c38da094!


sofar:
Yeah, that's pretty self-evident.

numberZero:
And the problem is, creation of VManip forces loading of all these blocks. That seems to be the reason of constant swapping on underpowered servers like SiE.

PR #408, opened 2018-01-27
Improve network parsing

numberZero:
Reduces parse time from ~6s to <1s for this construction (16³ nodes):
screenshot_20180128_020951
Internally, it reduces parsing complexity from O(n²) to O(n) (a bit heavier actually, due to reallocations; but still way faster than current one), so for larger networks the effect should be even more noticeable (although mapblock loading may reduce it considerably; depends).
@VanessaE you might be interested in this patch =)

### DIFF
--- a/technic/machines/switching_station.lua
+++ b/technic/machines/switching_station.lua
@@ -190,9 +190,13 @@ local get_network = function(sw_pos, pos1, tier)
 	local all_nodes = {}
 	local queue = {}
 	add_cable_node(all_nodes, pos1, network_id, queue)
-	for _, pos in ipairs(queue) do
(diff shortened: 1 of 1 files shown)


Ekdohibs:
Whaaaat how did I miss that? I'd like some testing (@VanessaE maybe you will want to on your server?), but once this is done, I'll merge this immediately! It explains a lot why building the network was so slow, which I didn't understand since I thought it was linear complexity ><.

VanessaE:
It's been put into effect on my servers.

Ekdohibs:
https://github.com/minetest-mods/technic/commit/86fb981b445bc63d564f7903a9a0d48b87a12763

PR #414, opened 2018-02-12
Update of substances > ore with lead and sulfur

tanmayameher:
Since Lead and Sulfur are missing from manual and they play an important role in making RE battery i thought it will be important to add them to Manual. Plus, Lead blocks got a place now in HV nuclear reactor layout.


bvs23bkv33:
thanks!

manual.md
```diff
@@ -42,8 +42,8 @@ of about +1.4.)  Positive elevations are above the reference plane and
 negative elevations below.  Because elevations are always described this
 way round, greater numbers when higher, we avoid the word "depth".
 
-The ores that matter in technic are coal, iron, copper, tin, zinc,
-chromium, uranium, silver, gold, mithril, mese, and diamond.
+The ores that matter in technic are coal, lead, sulfur, iron, copper, 
+tin, zinc, chromium, uranium, silver, gold, mithril, mese, and diamond.
```
numberZero: Strictly speaking, sulfur is not an ore. But neither diamond is.

manual.md
```diff
@@ -54,6 +54,17 @@ machines are electrically powered, but burning fuel remains a way to
 generate electrical power.  Coal is also used, usually in dust form, as
 an ingredient in alloying recipes, wherever elemental carbon is required.
 
+Lead is special ore and part of the Technic Mod. It is  found from elevation
++16 downwards, and its abundance decreases upto -128. After that there is
```
numberZero: Decreases?

numberZero:
Note that you can fully rewrite the branch history at any time. That is usually not necessary if PR is to be squashed (that’s the case for most PRs), but just for your info, it may be done using the commands:
```bash
# start history from master (not creating a branch yet)
git checkout --detach master

# list all commits in some-branch but not in master.
# --oneline and |cat are optional but convenient in this case IMO.
git log --oneline master..[some-branch] | cat

# add an existing commit to the current branch
git cherry-pick [some-commit-id]

# take a file from an existing commit
# don’t forget to commit such change afterwards.
git checkout [commit-or-branch-id] -- [file-name]

# reset a branch to the current HEAD
git checkout -B [branch-name]

# finally, push the reworked branch
# --force is used as this rewrites the branch history
# names are usually optional but you’d better specify them in force-push
git push --force [remote-name] [branch-name]
```

manual.md
tanmayameher: Yes, true. But, it is good to write some info on them. May be in a subsection as extra minerals or something?!

manual.md
tanmayameher: Sorry, i misunderstood the `clust_scarcity`, `clust_num_ores`, `clust_size` in `technic_worldgen/oregen.lua` . So, it seems, it stays the same as comparing with some other minerals. No variation in abundance.

numberZero: It increases, as far as I can understand.

manual.md
bvs23bkv33: sulfur can be grinded and diamond can not

tanmayameher: lol may be because diamond is the hardest known substance on earth xD
But, still what to do with diamond dust anyway. 

Sulfur is actually useful in organic compound area esp. like fertilizer, pharmaceuticals etc. since Sulfur reacts with most elements except few noble metals like gold, platinum and noble gases. 

Battery; strictly speaking, should use more of lead, copper, Lithium, Aluminum or other such elements. May be even distilled water like those in heavy duty lead acid battery.

manual.md
tanmayameher: Hmm... Yes. May be. I still has problem understanding them :)

`clust_num_ores` and `clust_size` remaining same for `clust_scarcity` variations. Other ores having similar value for those variables for instance, chromium written as 'no variation' in manual. So, are you sure it increases ?!

Although, in game it feels like it increases from ground, downwards to some point and after that its pretty much everywhere. 👍

numberZero: http://dev.minetest.net/minetest.register_ore

manual.md
numberZero: I think that’s not a problem for a mod. It’s not a textbook.

**OT**
Diamond dust is commonly used as an abrasive.
And on batteries: the simplest RE battery design I know consists of lead electrodes and sulfuric acid, and the simplest conventional battery I know consists of zinc and copper (or carbon) electrodes and salt water.

tanmayameher: Right 👍 

That's what i was trying to convey to bvs23bkv33.
This is a game. We don't need to be exactly IRL like. But, yes somewhere description of those two elements should be included for information purpose :)

Ah and thanks @numberZero for diamond dust usage info. I didn't know that. Now, i know one extra thing about diamond 👍

PR #415, opened 2018-02-13
Add Support for Ethereal NG plant life

cordial-cobra:
Adds in the ability of the chainsaw to cut down plant life generated by the Ethereal mod.
Power level was changed but still needs to be tested out to see whether it will cut down a whole redwood tree and use the whole charge without having any leftover. If it doesn't cut down the whole tree, aesthetically it will not look pleasing to the eye along with leaving nodes hanging out in the open.
Feedback appreciated!

### DIFF
--- a/technic/machines/switching_station.lua
+++ b/technic/machines/switching_station.lua
@@ -242,8 +242,8 @@ minetest.register_abm({
 
 			local poshash = minetest.hash_node_position(pos)
 
-			if not technic.redundant_warn.poshash then
-				technic.redundant_warn.poshash = true
+			if not technic.redundant_warn[poshash] then
+				technic.redundant_warn[poshash] = true
 				print("[TECHNIC] Warning: redundant switching station found near "..minetest.pos_to_string(pos))
 			end
 			return


technic/tools/chainsaw.lua
```diff
@@ -77,7 +78,35 @@ if minetest.get_modpath("moretrees") then
 		timber_nodenames["moretrees:apple_blossoms"]           = true
 	end
 end
-
+-- adds in support to cut down ethereal plant life
+if minetest.get_modpath("ethereal") then
+	timber_nodenames ["ethereal:redwood_trunk"]								= true
```
h-v-smacker: Extra whitespace before '['

technic/tools/chainsaw.lua
```diff
@@ -77,7 +78,35 @@ if minetest.get_modpath("moretrees") then
...
+		timber_nodenames["ethereal:redwood_leaves"]							= true
+		timber_nodenames["ethereal:yellowleaves"]   						= true
+		timber_nodenames["ethereal:mushroom"]								= true
+		timber_nodenames["ethereal:mushroom_pore"]							= true
+		timber_nodenames["ethereal:frost_leaves"]							= true
+		timber_nodenames["ethereal:yellowleaves"]							= true
+		timber_nodenames["ethereal:bananaleaves"]							= true
+		timber_nodenames["ethereal:orange_leaves"]							= true
+		timber_nodenames["ethereal:willow_twig"]							= true
+		timber_nodenames["ethereal:palmleaves"]								= true
```
h-v-smacker: You forgot the fruits. If you look at the chainsaw's code, you'll see that the fruits are usually listed alongside the leaves, so that they'd be harvested at the same time, allowing the player not to wait for their "natural" fall.

technic/tools/chainsaw.lua
```diff
@@ -77,7 +78,35 @@ if minetest.get_modpath("moretrees") then
...
+	timber_nodenames["ethereal:willow_trunk"]								= true
+	timber_nodenames["ethereal:big_tree"]									= true
+	timber_nodenames["ethereal:orange_tree"]								= true 
+
+--add support for ethereal leaves
+	if chainsaw_leaves then
+		timber_nodenames["ethereal:redwood_leaves"]							= true
+		timber_nodenames["ethereal:yellowleaves"]   						= true
+		timber_nodenames["ethereal:mushroom"]								= true
+		timber_nodenames["ethereal:mushroom_pore"]							= true
```
h-v-smacker: With the current tree traversal algorithm, neither acacia nor mushrooms are cut properly. The mushrooms will leave most of their caps floating in midair, just like the top of acacia crowns are left behind.

cordial-cobra: I've tried this out when I made the original change on a local game and it cut down the whole thing.

technic/tools/chainsaw.lua
cordial-cobra: Neglected this. Thanks for pointing that out.

technic/tools/chainsaw.lua
h-v-smacker: I did the same change at my fork of technic yesterday, and it didn't cut the whole thing until I changed the search area from 3x3 to 5x5.

technic/tools/chainsaw.lua
```diff
@@ -1,7 +1,8 @@
 -- Configuration
 
-local chainsaw_max_charge      = 30000 -- Maximum charge of the saw
--- Gives 2500 nodes on a single charge (about 50 complete normal trees)
+local chainsaw_max_charge      = 300000 -- Maximum charge of the saw
```
t4im: I understand that you want an entire redwood tree to be chainsawed in one go, but consider that this also affects all the other trees, even without ethereal.

Even right now the chainsaw can sometimes lead to smaller forests being devastated (even in civilized areas). Increasing this issue ten-fold just won't work well.

This rather needs changes to the traversal algorithm.
It may also help to make the algorithm aware of the tree type and limit traversal of nodes of the same type of tree and adapt charge relatively to that. This might also lead to somewhat cleaner results (at least in mixed forests).

Since that is probably out of scope of this PR i suggest to remove the increase.

GreenXenith:
bump

GreenXenith:
bump @cordial-cobra

auouymous:
This can be closed by the merged #570.

PR #418, opened 2018-03-15
pt_BR Translation Updated

Sires0:
afmachado Already made a translation to pt_BR, ik, but I made an updated one, with some more used words that have the same meaning and adapted to the way the translations are gonna be used

Also I didn't know afmachado already did this do I made it all from the zero xD Hard work lol


Sires0:
Not a single comment? Now there is one

numberZero:
Sorry but the project is virtually inactive currently. https://github.com/minetest-mods/technic/commits/master

SmallJoker:
Is your translation still up to date? And do the characters always display correctly?
The PR looks good, can be merged if it's tested.

Issue #419, opened 2018-03-26
Registering a new machine is impossible from another mod


LunaSquee:
Hello!
So I wanted to make a mod that would add more machines that work with the power network provided by technic, however I ran into a snag. The function `technic.register_base_machine` is hardcoded to prefix the node with `technic:` and thus cannot be called from another mod, as it will complain about there being an invalid prefix.

[File and line in question](https://github.com/minetest-mods/technic/blob/master/technic/machines/register/machine_base.lua#L147)

naturefreshmilk:
I registered a consumer successfully in my mod.
This worked for me:

```lua
if minetest.get_modpath("technic") then
	technic.register_machine("HV", "jumpdrive:engine", technic.receiver)
end
```

PR #420, opened 2018-03-30
Add mod name option to register base machine (#419)

LunaSquee:
Make it possible for other mods to use technic as an API for adding more machines compatible with technic networks.


SmallJoker:
Looks good, but needs to be documented somewhere.

PR #423, opened 2018-05-13
Use a:iter instead of for loops for sulfur generation

HybridDog:
The a:iter function is supposed to be used in such situations, it's faster than using for loops with a:index.


SmallJoker:
Test code:
```Lua
local total = 0
local count = 0
minetest.register_on_generated(function(minp, maxp, seed)

		if (c == c_lava or c == c_lava_flowing)
		and sulfur_noise:get_3d({x = x, y = z, z = z}) >= 0.4 then
			local t0 = os.clock()
			-- Old or new loop here
			local t1 = os.clock()
			total = total + (t1 - t0)
			count = count + 1

	if count > 0 then
		print("Took " .. math.ceil((total / count) * 1000 * 1000) .. " us")
	end
	vm:set_data(data)
```
Using mgv6 and LuaJIT.
Old loop: 180 µs
New loop: 88 µs
Definitely an improvement.

PR #424, opened 2018-05-13
Show EU power values more readable

HybridDog:
Add the EU_string helper function
In comparison to pretty_num it uses SI prefixes, adds "EU" (e.g. kEU) and rounds the number for readability

Before:
old

After:
new
(of course, the BB collected some power between taking the screenshots)

I've noticed that infotext is sometimes line-broken in minetest, which would puts the EU unit into the next line. Therefore I've changed the space to a non-breaking space character: ` `, are you fine with this?

### DIFF
--- a/technic/doc/api.md
+++ b/technic/doc/api.md
@@ -11,9 +11,12 @@ switching station handles the network activity.
 
 Helper functions
 ----------------
+* `technic.EU_string(num)`
+	* Converts num to a human-readable string with unit
+	* Use this function when showing players power values
 * `technic.pretty_num(num)`
-	* Converts the number `num` to a human-readable string.
-	* Use this function when showing players power values.
+	* Converts the number `num` to a human-readable string
+	* This function is currently unused by technic (legacy).
 * `technic.swap_node(pos, nodename)`
 	* Same as `mintest.swap_node` but it only changes the nodename.
 	* It uses `minetest.get_node` before swapping to ensure the new nodename
--- a/technic/helpers.lua
+++ b/technic/helpers.lua
@@ -21,6 +21,32 @@ function technic.pretty_num(num)
 end
 
 
+local prefixes = {[-24] = "y", [-21] = "z", [-18] = "a", [-15] = "f",
+	[-12] = "p", [-9] = "n", [-6] = "µ", [0] = "", [-3] = "m", [3] = "k", [6] = "M",
+	[9] = "G", [12] = "T", [15] = "P", [18] = "E", [21] = "Z", [24] = "Y"}
+local digits = 4  -- shouldn't be less than 4
+function technic.EU_string(num)
+	if num == 0
+	or num * 0 ~= num * 0 then
+		-- ±0, ±inf or ±nan
+		return tostring(num)
+	end
+	local b = math.floor(math.log(num) / math.log(10) +0.000001)
+	b3 = math.floor((b - math.sign(b)) / 3) * 3
+	if math.abs(b) < digits-1 then
+		b = 0
+	else
+		b = b - (digits-1) * math.sign(b)
+	end
+	if not prefixes[b3] then
+		-- not likely going to happend
+		return tostring(num) .. " EU"
+	end
+	num = math.floor(num / 10^b +.5) * 10^(b - b3)
+	return num .. " " .. prefixes[b3] .. "EU"
+end
+
+
 --- Same as minetest.swap_node, but only changes name
 -- and doesn't re-set if already set.
 function technic.swap_node(pos, name)
--- a/technic/helpers.lua
+++ b/technic/helpers.lua
@@ -40,10 +40,10 @@ function technic.EU_string(num)
 	end
 	if not prefixes[b3] then
 		-- not likely going to happend
-		return tostring(num) .. " EU"
+		return tostring(num) .. " EU"
 	end
 	num = math.floor(num / 10^b +.5) * 10^(b - b3)
-	return num .. " " .. prefixes[b3] .. "EU"
+	return num .. " " .. prefixes[b3] .. "EU"
 end
 
 
--- a/technic/helpers.lua
+++ b/technic/helpers.lua
@@ -26,9 +26,8 @@ local prefixes = {[-24] = "y", [-21] = "z", [-18] = "a", [-15] = "f",
 	[9] = "G", [12] = "T", [15] = "P", [18] = "E", [21] = "Z", [24] = "Y"}
 local digits = 4  -- shouldn't be less than 4
 function technic.EU_string(num)
-	if num == 0
-	or num * 0 ~= num * 0 then
-		-- ±0, ±inf or ±nan
+	if num * 0 ~= num * 0 then
+		-- ±inf or ±nan
 		return tostring(num)
 	end
 	local b = math.floor(math.log(num) / math.log(10) +0.000001)
--- a/technic/machines/LV/solar_panel.lua
+++ b/technic/machines/LV/solar_panel.lua
@@ -35,7 +35,8 @@ local run = function(pos, node)
 		local charge_to_give = math.floor((light + pos1.y) * 3)
 		charge_to_give = math.max(charge_to_give, 0)
 		charge_to_give = math.min(charge_to_give, 200)
-		meta:set_string("infotext", S("@1 Active (@2 EU)", machine_name, technic.pretty_num(charge_to_give)))
+		meta:set_string("infotext", S("@1 Active (@2)", machine_name,
+			technic.EU_string(charge_to_give)))
 		meta:set_int("LV_EU_supply", charge_to_give)
 	else
 		meta:set_string("infotext", S("%s Idle"):format(machine_name))
@@ -54,7 +55,7 @@ minetest.register_node("technic:solar_panel", {
 	active = false,
 	drawtype = "nodebox",
 	paramtype = "light",
-	is_ground_content = true,	
+	is_ground_content = true,
 	node_box = {
 		type = "fixed",
 		fixed = {-0.5, -0.5, -0.5, 0.5, 0, 0.5},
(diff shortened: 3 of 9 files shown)


technic/helpers.lua
```diff
@@ -21,6 +21,31 @@ function technic.pretty_num(num)
 end
 
 
+local prefixes = {[-24] = "y", [-21] = "z", [-18] = "a", [-15] = "f",
+	[-12] = "p", [-9] = "n", [-6] = "µ", [0] = "", [-3] = "m", [3] = "k", [6] = "M",
+	[9] = "G", [12] = "T", [15] = "P", [18] = "E", [21] = "Z", [24] = "Y"}
```
numberZero: Why do you step by 3 and not by 1?

technic/helpers.lua
```diff
@@ -21,6 +21,31 @@ function technic.pretty_num(num)
...
+local digits = 4  -- shouldn't be less than 4
+function technic.EU_string(num)
+	if num * 0 ~= num * 0 then
+		-- ±inf or ±nan
```
numberZero: Ugh, Lua lacks `isfinite`...

technic/helpers.lua
```diff
@@ -21,6 +21,31 @@ function technic.pretty_num(num)
...
+	local b = math.floor(math.log(num) / math.log(10) +0.000001)
+	b3 = math.floor((b - math.sign(b)) / 3) * 3
+	if math.abs(b) < digits-1 then
+		b = 0
+	else
+		b = b - (digits-1) * math.sign(b)
+	end
+	if not prefixes[b3] then
+		-- not likely going to happend
+		return tostring(num) .. " EU"
```
numberZero: `string.format("%e EU", num)`

technic/helpers.lua
```diff
@@ -21,6 +21,31 @@ function technic.pretty_num(num)
...
+local prefixes = {[-24] = "y", [-21] = "z", [-18] = "a", [-15] = "f",
+	[-12] = "p", [-9] = "n", [-6] = "µ", [0] = "", [-3] = "m", [3] = "k", [6] = "M",
+	[9] = "G", [12] = "T", [15] = "P", [18] = "E", [21] = "Z", [24] = "Y"}
+local digits = 4  -- shouldn't be less than 4
+function technic.EU_string(num)
+	if num * 0 ~= num * 0 then
+		-- ±inf or ±nan
+		return tostring(num)
+	end
+	local b = math.floor(math.log(num) / math.log(10) +0.000001)
```
numberZero: But Lua does have `log10`.
BTW, what if `num` is exactly zero?

technic/helpers.lua
```diff
@@ -21,6 +21,31 @@ function technic.pretty_num(num)
...
+	b3 = math.floor((b - math.sign(b)) / 3) * 3
+	if math.abs(b) < digits-1 then
+		b = 0
+	else
+		b = b - (digits-1) * math.sign(b)
+	end
```
numberZero: Some magic...

technic/helpers.lua
```diff
@@ -21,6 +21,31 @@ function technic.pretty_num(num)
...
+	if not prefixes[b3] then
+		-- not likely going to happend
+		return tostring(num) .. " EU"
+	end
+	num = math.floor(num / 10^b +.5) * 10^(b - b3)
+	return num .. " " .. prefixes[b3] .. "EU"
```
numberZero: More magic...
Why not simply something like
```lua
b = math.floor(math.log10(num) / 3 + 0.001)
return string:format("%.2f %sEU", num / 1000^b, prefixes[b])
```
?

technic/helpers.lua
HybridDog: yes, the comment explains what is done here

HybridDog:
> why do you add new function instead of updating pretty_num?

When I wrote the new function I expected pretty_num to be used not only for power values but also for something different, but it turned out to being always used for power.
I thought I'd keep the new function name because pretty_num does not attach "EU" and the function name is therefore quite hazy.

technic/helpers.lua
HybridDog: I thought this makes the code a bit easier to understand, I've changed it to 1.

technic/helpers.lua
HybridDog: I've used `"%.4g EU"` because l noticed that this is also used when num is 0.

technic/helpers.lua
HybridDog: > But Lua does have log10.

Thanks, I didn't know this because l rarely use it…

> BTW, what if num is exactly zero?

See above.

technic/helpers.lua
HybridDog: For readability, it shows e.g. 4398 kEU instead of 4.298 MEU but to avoid showing 1001 mEU, as an edge case 1.001 EU is shown.

technic/helpers.lua
HybridDog: This doesn't incorporate the edge case, see above.

technic/helpers.lua
```diff
@@ -21,6 +21,30 @@ function technic.pretty_num(num)
...
+	else
+		-- as special case, avoid showing e.g. 1100 mEU instead of 1.1 EU
+		pref_i = math.floor(b / 3)
+	end
+	if not prefixes[pref_i] then
+		-- This happens for 0, nan, inf, very big values, etc.
+		return string.format("%.4g EU", num)
+	end
+
+	return string.format("%.4g %sEU", num / 10^(pref_i * 3), prefixes[pref_i])
```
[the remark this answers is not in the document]
HybridDog: What do you mean?
Specifying 4 digits is intended.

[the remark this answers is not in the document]
numberZero: Sorry for confusing you. I understood that that’s how `%g` works, but I didn’t knew that before. It appears to be really handy)

technic/helpers.lua
```diff
@@ -21,6 +21,30 @@ function technic.pretty_num(num)
...
+	-- the small number added is due to floating point inaccuracy
+	local b = math.floor(math.log10(math.abs(num)) +0.000001)
+	local pref_i
+	if b ~= 0 then
+		-- b is decremented by 1 to avoid a single digit with many decimals,
+		-- e.g. instead of 1.021 MEU, 1021 kEU is shown
+		pref_i = math.floor((b - 1) / 3)
+	else
+		-- as special case, avoid showing e.g. 1100 mEU instead of 1.1 EU
+		pref_i = math.floor(b / 3)
```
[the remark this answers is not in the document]
HybridDog: fixed

HybridDog:
> You could make the function append the Si prefix but not the unit tag, like 4.40 M. I would even prefer to see such function, so that it could be used for Power Units (PU, EU/s), for example (could be displayed by the supply converter, the power monitor, etc.)

I've changed the pretty_num function now, EU_string appends "EU" to the pretty_num function's result.

> So if you really prefer the heuristic you describe, I take back my objections

I prefer it because X.XXX numbers (X is digit) are avoided, XXXX is easier to understand in my opinion.

technic/helpers.lua
```diff
@@ -1,23 +1,33 @@
...
+		-- b is decremented by 1 to avoid a single digit with many decimals,
+		-- e.g. instead of 1.021 MEU, 1021 kEU is shown
+		pref_i = math.floor((b - 1) / 3)
+	else
+		-- as special case, avoid showing e.g. 1100 mEU instead of 1.1 EU
+		pref_i = 0
+	end
+	if not prefixes[pref_i] then
+		-- This happens for 0, nan, inf, very big values, etc.
+		return string.format("%.4g EU", num)
```
numberZero: er, `EU` is still here?

technic/helpers.lua
```diff
@@ -1,23 +1,33 @@
...
 	end
-	digit_sep_esc = sep
+
+	return string.format("%.4g %sEU", num / 10^(pref_i * 3), prefixes[pref_i])
```
numberZero: and here

technic/helpers.lua
```diff
@@ -1,23 +1,33 @@
...
 
-function technic.pretty_num(num)
-	local str, k = tostring(num), nil
-	repeat
-		str, k = str:gsub("^(-?%d+)(%d%d%d)", "%1"..digit_sep_esc.."%2")
-	until k == 0
-	return str
+-- used to display power values
+function technic.EU_string(num)
+	return technic.pretty_num(num) .. "EU"
```
numberZero: ...but added again here

technic/doc/api.md
```diff
@@ -11,9 +11,12 @@ switching station handles the network activity.
 
 Helper functions
 ----------------
+* `technic.EU_string(num)`
+	* Converts num to a human-readable string (see pretty_num)
+	  and adds the `EU` unit
+	* Use this function when showing players power values
```
numberZero: EU stands for *Energy* Unit, not *Power* Unit (which could be EU/s). Sadly technic uses EUs for both energy and power, which is incorrect.

HybridDog:
> Sadly technic uses EUs for both energy and power, which is incorrect.

EU/s is incorrect because the energy production speed depends on server lag.
#
I've fixed the EUEU mistake and added a constant digit count setting.

HybridDog:
`%.4g` emits zeros on the right, it shows e.g. "15" instead of "15.00"

numberZero:
Sorry, when I was talking about constant digit count, I didn’t know the exact `%g` behavior. It completely satisfies me.
So, I’m neutral on 78084a8, and :+1: on everything else. Sadly I’m only a contributor, and can’t merge anything...

SmallJoker:
This PR improves the readability a lot - I'm happy to see this change. However, could you please add unittests for `technic.pretty_num()` to ensure it outputs the strings as expected?

HybridDog:
I've added the unittests.
I've also disallowed showing `-0`.

PR #434, opened 2018-08-03
Fix bug where timers stop working when frames move

### LINKED ISSUE #432: Frames are very buggy with components from mesecons (blocks get stuck in weird states)
I tried to build a moving platform.

But noticed that the interaction between mesecons and frames is very buggy.

One simple way to replicate it is to place a frame motor for a moving platform, and put a button next to it. When you press the button, it will get stuck in the on state, and the platform will move endlessly.

The same can be said about mesecons and luacontrollers. Mesecons will sometimes get stuck in the on state when platforms move, and luacontrollers might also get stuck in an internal state when it shouldn't.

I assume that this is some kind of race condition, where if the mesecon components update while the frame platform is moving, all the updates will be reset when the platform has finished moving.

I don't know how these mods work internally. But if there was some way to disable mesecons (or any other block) from updating while moving the frames, that might solve the issue.

ryvnf:
This fixes #432 for buttons, blinky plants, delayers and other nodes which rely on timers to update.

Before this commit, if you place a button which activate a frame which moves the button the frame is on, the button will get stuck in its pressed state endlessly. After this commit, it will return to its unpressed state when the timer runs out.

This does not fix components from mesecons which rely on mod internal timers. Which include Lua Controllers, and Delayers. Fixing that would not be easy, because these internal timers rely on the node position which might have changed after the frames move. Both these components will sometimes work with frames, and sometimes get stuck in some state because of this problem, so they are unreliable to use. But this problem should probably be fixed in mesecons instead, as it is a problem with mesecons internal architecture.


technic/machines/other/frames.lua
```diff
@@ -16,34 +16,34 @@ local function get_face(pos,ppos,pvect)
 		local t=(-0.5-ppos.x)/pvect.x
 		local y_int=ppos.y+t*pvect.y
 		local z_int=ppos.z+t*pvect.z
-		if y_int>-0.45 and y_int<0.45 and z_int>-0.45 and z_int<0.45 then return 1 end 
+		if y_int>-0.45 and y_int<0.45 and z_int>-0.45 and z_int<0.45 then return 1 end
```
numberZero: Whitespace cleanup is a good thing, although it interferes with `git blame`.

technic/machines/other/frames.lua
```diff
@@ -133,6 +143,8 @@ local function move_nodes_vect(poslist,vect,must_not_move,owner)
 		minetest.set_node(npos, n.node)
 		local meta = minetest.get_meta(npos)
 		meta:from_table(n.meta)
+		local timer = minetest.get_node_timer(npos)
+		timer:set(n.timer.timeout, n.timer.elapsed)
```
numberZero: What if the timer wasn’t started?

SmallJoker: When the timer is inactive, it's not present in the mapblock timer list, thus the values `0`, `0` are returned (https://github.com/minetest/minetest/blob/1992db1/src/nodetimer.h#L71). However, it seems to be triggered instantly if there's a callback available. Nice catch.

ryvnf:
Nice! Didn't think about the no timer problem. I'll fix that tomarrow.

My editor automatically removed whitespace. Should that be in a seperate commit so it is more tracable?

SmallJoker:
@ryvnf For my part it's fine to keep the whitespace changes since some lines need some style cleanups anyway.

ryvnf:
I fixed the things you mentioned and made the file more readable. It is split in two commits, one for the bugfix, and one for the clean up.

The cleanup commit has a very long diff (about one thousand lines). I have reviewed every single line and tested the updated code on a moving mining machine.

Previously as you may have noticed, the code contained unnecessary parenthesis in many places. It also contained very little whitespace, and very long lines which made it hard to read in many places. I used regular expressions, macros and some hand work to make it more readable and consistent with the styles of other files in Technic.

Even though I have tested and reviewed it, it might be a good idea for someone to take their time to review the diff. Just to be sure. The more eyes on the code, the better.

technic/machines/other/frames.lua
```diff
@@ -518,102 +578,124 @@ minetest.register_on_dignode(function(pos, node)
...
-		if not(pos_in_list(c,pos1)) and nodename~="air" and
-		(minetest.registered_nodes[nodename].frames_can_connect==nil or
-		minetest.registered_nodes[nodename].frames_can_connect(pos1,vect)) then
-			c[#(c)+1]=pos1
-			if minetest.registered_nodes[nodename].frame==1 then
-				local adj=minetest.registered_nodes[nodename].frame_connect_all(nodename)
-				connected(pos1,c,adj)
+		if not pos_in_list(c, pos1) and nodename ~= "air" and
+		(minetest.registered_nodes[nodename].frames_can_connect == nil or
+		minetest.registered_nodes[nodename].frames_can_connect(pos1, vect)) then
```
SmallJoker: Would you please be so nice and adjust the indents here as well?

technic/machines/other/frames.lua
```diff
@@ -889,54 +987,54 @@ minetest.register_node("technic:template_motor",{
 minetest.register_craft({
 	output = 'technic:frame_111111',
 	recipe = {
-		{'',			'default:stick',	''},
-		{'default:stick',	'technic:brass_ingot',	'default:stick'},
-		{'',			'default:stick',	''},
+		{ '',			'default:stick',	'' },
```
SmallJoker: Mixing of dynamic size tabulators and regular characters. Spaces should be used to keep the recipe items lined-up for all tabulator sizes.
EDIT: Same for the changes below.

ryvnf:
I agree with all your comments and have amended the commit according to them.

Changed the commit title as well to reflect the further changes.

PR #437, opened 2018-08-18
Add "creative mode" option

cheapie:
This does not change the default behavior, but if "creative_mode" is set to true in technic.conf, this completely disables power distribution and causes the following to happen, regardless of whether there is a switching station connected:
* Switching Stations and Power Monitors become inert
* Generators placed on top of a suitable cable will run (consume fuel if applicable, update infotext, and so on)
* Batteries placed on top of a suitable cable will charge at maximum speed and remain fully charged
* Machines that consume energy and are placed on top of a suitable cable will run normally, regardless of whether any power is present
* Power distribution logic will not run and networks will not be calculated

This also includes an "enable_producers" option, added at the request of a server owner (I'd be open to a better name for this one). The default is true, but if set to false it skips registering all generator/producer nodes. It is intended for use in conjunction with creative mode but will work, albeit somewhat pointlessly, regardless.

### DIFF
--- a/technic/machines/creative.lua
+++ b/technic/machines/creative.lua
@@ -1,0 +1,41 @@
+local S = technic.getter
+
+minetest.register_abm({
+	nodenames = {"group:technic_lv", "group:technic_mv", "group:technic_hv"},
+	label = "Run Machines",
+	interval   = 1,
+	chance     = 1,
+	action = function(pos,node)
+		local meta = minetest.get_meta(pos)
+		local pos1 = {x=pos.x, y=pos.y-1, z=pos.z}
+		local tier = technic.get_cable_tier(minetest.get_node(pos1).name)
+		local meta = minetest.get_meta(pos)
+		if not tier then
+			meta:set_int("active", 0)
+			return
+		end
+		meta:set_int("active", 1)
+		meta:set_int("LV_EU_input", meta:get_int("LV_EU_demand"))
+		meta:set_int("MV_EU_input", meta:get_int("MV_EU_demand"))
+		meta:set_int("HV_EU_input", meta:get_int("HV_EU_demand"))
+		local nodedef = minetest.registered_nodes[node.name]
+		if nodedef and nodedef.technic_run then
+			nodedef.technic_run(pos, node)
+		end
+	end,
+})
+
+minetest.register_lbm({
+	nodenames = {"technic:switching_station", "technic:power_monitor"},
+	name = "technic:update_infotext",
+	label = "Update switching station / power monitor infotext",
+	run_at_every_load = true,
+	action = function(pos, node)
+		local meta = minetest.get_meta(pos)
+		if node.name == "technic:switching_station" then
+			meta:set_string("infotext", S("Switching Station"))
+		elseif node.name == "technic:power_monitor" then
+			meta:set_string("infotext", S("Power Monitor"))
+		end
+	end,
+})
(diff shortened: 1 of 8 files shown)


technic/machines/creative.lua
```diff
@@ -0,0 +1,40 @@
...
+		local nodedef = minetest.registered_nodes[node.name]
+		if nodedef and nodedef.technic_run then
+			nodedef.technic_run(pos,node)
+		end
+	end,
+})
+
+minetest.register_lbm({
+	nodenames = {"technic:switching_station","technic:power_monitor"},
+	name = "technic:update_infotext",
```
numberZero: `run_at_every_load`?

technic/machines/init.lua
```diff
@@ -13,3 +13,8 @@ dofile(path.."/supply_converter.lua")
 
 dofile(path.."/other/init.lua")
 
+if technic.config:get_bool("enable_creative_mode") then
```
numberZero: It should also check whether server itself runs in the creative mode.

technic/machines/creative.lua
```diff
@@ -0,0 +1,40 @@
+local S = technic.getter
+
+minetest.register_abm({
+	nodenames = {"group:technic_lv","group:technic_mv","group:technic_hv"},
+	label = "Run Machines",
+	interval   = 1,
+	chance     = 1,
+	action = function(pos,node)
+		local meta = minetest.get_meta(pos)
+		local pos1 = {x=pos.x,y=pos.y-1,z=pos.z}
```
numberZero: Comma **must** be followed by whitespace.

cheapie: That line was actually copy/pasted from existing switching station code, but I'll go ahead and change that in a moment.

technic/machines/creative.lua
numberZero: Creative mode is just an option which can be switched on and off as many times as world owner desires.

technic/machines/init.lua
numberZero: Or rename the option to `creative_mode`, as it *forces* that, not merely *allows*.

> possible to mix and match

That is always possible as `technic` is free.

cheapie: OK, option name changed.

technic/machines/creative.lua
cheapie: Yep, and the LBM runs once each time it is switched on. I have confirmed that it works as intended even when repeatedly switching back and forth.

numberZero: From the documentation:
```lua
        run_at_every_load = false,
        -- Whether to run the LBM's action every time a block gets loaded,
        -- and not just for blocks that were saved last time before LBMs were
        -- introduced to the world.
```
So you seem to rely on undocumented behavior. minetest/minetest#7660

cheapie: This is how LBMs were explained to me when they were first introduced, but I'll go ahead and add this. I suppose it can always be removed later if the docs are improved.

numberZero: What would suit better IMO is minetest/minetest#5303, but that is not in MT yet ([PR#5653](https://github.com/minetest/minetest/pull/5653)).

SmallJoker:
People who use technic in creative mode want to test whether their circuit idea might work. I don't see any benefits in adding a creative_mode setting, also because the machine outputs can be obtained by the creative inventory.

cheapie:
We've actually managed to achieve what we were trying to do (lag reduction, mostly) by other means, so I'm closing this. I'll leave my repository with this code up for a while in case there's any interest in it.

PR #441, opened 2018-08-24
Add a new type of cable clip

VanessaE:
Uses Zeg9's steel mod's "steel:strut_mount" or streetsmod's "streets:steel_support" as the base/blocky part, with an overlay copied from the steel mod to make it look like the clip is fixed to the strut with a steel band.
    
Textures will adjust to match whichever mod is installed, preferring the steel mod.
    
Can be crafted from any of these standard/shaped recipes:
```
 - f -        - f d        - f d
 - m -   or   - m -   or   - s i
 - - -        - - -        - - -
```
f = fencepost-shaped clip
m = steel mod strut with mount
s = steel mod strut without mount, or streets mod steel support
i = default steel ingot
d = dye (optional, see below)
    
If the user has Unified Dyes (commit 2a816534 or later), the clip can be dyed (well, the white part anyway :-) ), either by adding a portion of dye as above, or by crafting an existing clip + dye. Uses "colorwallmounted" mode for 32-color support.

IMPORTANT: This must be applied on top of #438 or in lieu of it.


VanessaE:
rebased.

extranodes/init.lua
```diff
@@ -146,6 +146,33 @@ local iclipfence_def = {
...
+local sclip_tex = {
+	"technic_insulator_clip.png",
+	{ name = "strut.png^steel_strut_overlay.png", color = "white" },
+	{ name = "strut.png", color = "white" }
+}
+
+if minetest.registered_nodes["streets:steel_support"] then
+	sclip_tex = {
+		"technic_insulator_clip.png",
+		{ name = "streets_support.png^technic_steel_strut_overlay.png", color = "white" },
```
SmallJoker: `2018-08-27 23:42:27: ERROR[Main]: generateImage(): Could not load image "streets_support.png" while building texture; Creating a dummy image`
The issue is that the steel mod adds an alias when this node was not found:
> minetest.register_alias("streets:steel_support","steel:strut")

I recommend checking whether the mods are loaded, either alone or additionally to the node check.

extranodes/init.lua
```diff
@@ -181,6 +216,39 @@ minetest.register_craft({
...
+		recipe = {
+			{"technic:insulator_clip_fencepost"},
+			{"steel:strut"},
+			{"default:steel_ingot"}
+		}
+	})
+
+end
+
+if minetest.registered_nodes["streets:steel_support"] then
```
SmallJoker: Same issue here

VanessaE: Fixed both.  Note that since steel mod has the strut with steel band PR already merged, I dropped one of the conditionals (since it'd be redundant).

VanessaE: Also opened a PR against steel mod to make a similar change.

VanessaE: Changed recipes to be consistent with how the base strut is crafted

PR #443, opened 2018-09-09
Fix connect_sides of active fuel-fired generators

12Me21:
Fuel-fired generators can output power on the bottom, back, left, and right.
Inactive generators had connect_sides correctly set to {"bottom", "back", "left", "right"}, but *active* generators only had {"bottom"} for some reason.
This caused generators to appear to disconnect from cables after being turned on.


12Me21:
So apparently generators actually connect on ALL sides, and this is an entirely separate bug. I'm not sure what sides they are SUPPOSED to output power from (apparently just the bottom?).

PR #446, opened 2018-09-17
Fix cable plate placement

12Me21:


12Me21:
bump

technic/machines/register/cables.lua
```diff
@@ -201,6 +201,19 @@ function technic.register_cable(tier, size)
 		def.node_box["connect_"..notconnects[i]] = nil
 		if i == 1 then
 			def.on_place = function(itemstack, placer, pointed_thing)
+				
+				-- Quick fix to prevent placing cables in protected areas and overwriting other nodes with cables
+				local name = placer:get_player_name()
+				if minetest.is_protected(pointed_thing.above, name) then
+					minetest.record_protection_violation(pos, name)
```
t4im: Since you don't check, that you got the right `pointed_thing` first (after `buildable_to` checks) and because that might be out of scope for a quickfix, you might want to replace the recording with a logging statement for the time being.

Otherwise you risk people getting kicked/jailed/harmed for trying to replace flowers.

12Me21:
(I'm pretty sure the protection_bypass privilege is handled automatically by `minetest.is_protected` now.)
But anyway, I've decided to just use `minetest.item_place` to handle placing the node, because it's a lot safer than checking for all possible special cases and using `set_node`.
It's pretty simple to override the node that is placed by passing a temporary itemstack and then checking to see if any items were removed.

technic/machines/register/cables.lua
```diff
@@ -117,6 +117,15 @@ local function clear_networks(pos)
 	end
 end
 
+local function item_place_override_node(itemstack, placer, pointed, node)
+	-- Call the default on_place function with a fake itemstack
+	local temp_itemstack = ItemStack(node.name)
+	temp_itemstack = minetest.item_place(temp_itemstack, placer, pointed, node.param2) or temp_itemstack
```
SmallJoker: Line is quite long. Needs breaks.

technic/machines/register/cables.lua
```diff
@@ -228,11 +237,7 @@ function technic.register_cable(tier, size)
 						num = xyz[((fine_pointed[bigger] < 0 and "-") or "") .. bigger]
 					end
 				end
-				minetest.set_node(pointed_thing.above, {name = "technic:"..ltier.."_cable_plate_"..num})
-				if not (creative and creative.is_enabled_for(placer)) then
-					itemstack:take_item()
-				end
-				return itemstack
+				return item_place_override_node(itemstack, placer, pointed_thing, {name = "technic:"..ltier.."_cable_plate_"..num})
```
SmallJoker: Line is quite long. Needs breaks

technic/machines/register/cables.lua
```diff
@@ -117,6 +117,15 @@ local function clear_networks(pos)
 	end
 end
 
+local function item_place_override_node(itemstack, placer, pointed, node)
+	-- Call the default on_place function with a fake itemstack
+	local temp_itemstack = ItemStack(node.name)
```
SmallJoker: Why do you strip the metadata? This could be relevant for colored nodes (or similar) when this function is used for a different item.

12Me21:
Well it's not 1980 anymore but yeah I'll fix that

technic/machines/register/cables.lua
12Me21: Which itemstack? I'm not sure how the metadata could be lost here.

PR #451, opened 2018-10-31
Use my basic_materials mod where possible

VanessaE:


numberZero:
Technic is large and self-contained. Depending on something unrelated is not good IMO.
I might be fine with soft dependency, though.

P.S. The spool disappears in the suggested craft recipes.

VanessaE:
Technic's never really been self-contained.  Technic itself depends on pipeworks and technic_worldgen already, and some parts require mesecons, moreores, or other mods to work properly.

VanessaE:
Oh as for wire, I figured in the recipes where the spool is not returned, it's treated as "consumed".  Like, a machine that needs wire surely needs the spool's plastic as insulation for that wire. :smiley:

VanessaE:
Technic's never really been self-contained. Technic itself depends on pipeworks (which depends on Mesecons, basic_materials, and a couple of others), and technic_worldgen already.  Some parts require other mods to work properly.  For example, you can't make silver wire and one of the energy crystals without Moreores (for the mithril in the latter case).

As for the spools of wire, I figured in the recipes where the empty spool is not returned, it's treated as "consumed". Like, a machine that needs wire surely needs the spool's plastic as insulation for that wire. :smiley:

DS-Minetest:
>As for the spools of wire, I figured in the recipes where the empty spool is not returned, it's treated as "consumed". Like, a machine that needs wire surely needs the spool's plastic as insulation for that wire.

Sure, and when clothes get sewn, the empty coils are used to make the knobs. :neutral_face:
Or the trash can "consumes" it.

VanessaE:
It's Minetest.  You can't begrudge me a little "magic" :stuck_out_tongue:

VanessaE:
There, replacements added.  I think I got all of them anyway.

numberZero:
@VanessaE Your mod is unsuitable for `technic` IMO, not to mention that it is not ready. Metals are OK, plastic is acceptable, but... motor? How could it be considered basic *material?* And energy crystal, as provided by `basic_materials`, is obviously something magical and not technical (`technic` needs a better recipe for its own crystals ofc).

DS-Minetest:
The diamond blocks make tools more wealthy [`basic_materials:energy_crystal_simple` is much cheaper](https://gitlab.com/VanessaE/basic_materials/blob/master/electrical-electronic.lua#L62). And a torch is probably not the best thing to store power.

Can't `basic_materials` instead simply make aliases to support `technic`?

VanessaE:
> make tools more wealthy

In vanilla minetest_game, diamond is not very hard to acquire, and my energy crystal uses diamond (and mese) anyway.

> store power

Technic energy crystals all have batteries in their recipes, so clearly those are the energy storage.  The torch in my energy crystal is actually kind of a legacy item -- it just represents a light source, as those crystals were originally just used for Home Decor's glow lights, from a time when torches were the only light source in minetest_game (not counting lava).

Think of my energy crystal like a Khyber crystal in Star Wars -- it's a regulator/conditioner, not the energy source.

In any case, it sounds like your real complaint is the recipe for my simple energy crystal, not the fact that I'm using it in technic's recipes for the big energy crystals.

VanessaE:
> make aliases to support technic

Not possible, because of the way Minetest handles them.  In order for an alias to be useful, either there must not be an item with the same name, or I'd have to use `register_alias_force`, which would require my mod depending on technic instead of the opposite, which would eventually result in a circular dependency somewhere (A depends on B depends on C depends on A...).

luk3yx:
The latest `pipeworks` depends on `basic_materials`, and technic requires `pipeworks` to be installed to work. Why should technic not take advantage of a mod that is most likely installed?

Another solution would be to create a `basic_materials.lua` file that registers `basic_materials` items that are used, then `basic_materials` could be an optional dependency. However, technic already indirectly depends on `basic_materials`, and this file would be unused unless `pipeworks` is outdated.

Ekdohibs:
As `pipeworks` already depends on `basic_materials`, it seems completely normal to use it instead of technic-specific items, so :+1:.

Ekdohibs:
Since `pipeworks` does depend on `basic_materals` and `technic` depends on `pipeworks`, having an optional dependency doesn't really make sense. Thus, I am merging this pull request.

PR #452, opened 2018-11-04
New HV-machines

upsala:
Electric-furnance, grinder, alloy-furnance, extractor and centrifuge as HV-Versions, with minor code-changes. Images are taken from the MV-version and are colored in blue.


t4im:
One example on how to not destroy gameplay could be for example to add HV versions of some machines that are able to process blocks, but are *not* able to process dust or ore. A Trade-off that still requires MV machines to operate, and that can be made slower per item to avoid dominance.

upsala:
But that's also true for the MV-Machines. If I have MV-machines, the LV-machines also getting obsolete. And removing functions from HV-machines, which MV has, doesn't sound useful to me. But this code reduce server-load, because you need lesser machines.

t4im:
That pending "LV obsoletion" has been a occasional critique point as well among players.
Which is why we shouldn't double down on it with MV. :-)

At least LV still has the "beginner tier" usage and MV adds the automation feature.
This reappearing HV machine idea tries to obsolete MV entirely without adding anything new.

> And removing functions from HV-machines, which MV has, doesn't sound useful to me.

Don't think so much like a player that tries to solve the challenge by coding and a little more like a developer that tries to create that challenge. We are both, which requires us to make that role distinction consciously.

The player retains the option to build both machine tiers. So by distributing those features you are *more than doubling* the game play mechanics. I.e. now the player has to build both MV and HV machine setups and even combine them! That's emergent gameplay right there. :-D

It requires completely new recipes to be added! Not just bigger versions of the old ones. New unique crafting chains can also enrich the gameplay here.

> But this code reduce server-load, because you need lesser machines.

If you want to work on performance, there are much more critical areas to address, that do not trade-off gameplay for it. Power-Network traversal comes to mind.

Please understand that i'm not criticising you or your PR, both are fine.
It's this reoccurring idea itself, that is lacking in its concept.
There is clearly a wish and need for "something" in this area, but it should not be solved like this.

h-v-smacker:
In my fork that we use over at LinuxForks server, I just introduced three HV machines: furnace, grinder, and compressor. They are the only ones that are involved in building material processing. Which leaves MV tier with alloying, separation, threshing (something you don't have here), and extracting. And the LV tier still offers the CNC which is very important for architecture, the music player (if someone uses it), and led lighting (again, something you don't have). This way all three tiers have their unique specialties while also participating in vertical upgrade route.

thomasrudin:
I for one like the idea of new machines, how about hiding the "overpowered" ones under a setting (those in technic/config.lua are a pull-request of its own...)
```
technic.op_machines = true
```

just my two cents...

h-v-smacker:
> On a related note: A few of your ideas in your technic branch are quite awesome, how about a few pull-requests...

I already tried to get something into upstream technic, it went nowhere. I see no point in making PRs which nobody wants.

SmallJoker:
I agree with t4im that this would make MV machines obsolete unless there's features that would make either a better choice, depending on the situation (e.g. tube compatibility, upgrades or speed).

Closing: abandoned for a long time and a rebase is needed as well.

PR #457, opened 2018-11-25
MV-Freezer

upsala:
Ok, another try. An MV-Freezer which handles the following recipes:

Bucket of Water => Snowblock & Bucket
Bucket of River-Water => Snowblock & Bucket
Dirt => Dirt with Snow


12Me21:
Wouldn't a block of ice make more sense?

12Me21:
I guess...
But realistically, ice should be easier to produce than snow. Ice is just frozen water while snow is made of tons of tiny particles of ice mixed with air.

upsala:
That's right, but we have no default recipe to create snow.

technic/machines/MV/freezer.lua
```diff
@@ -0,0 +1,12 @@
+-- MV freezer
+
+minetest.register_craft({
+    output = 'technic:mv_freezer',
+    recipe = {
+        {'technic:stainless_steel_ingot', 'technic:stainless_steel_ingot',  'technic:stainless_steel_ingot'},
+        {'pipeworks:tube_1',              'technic:mv_transformer', 'pipeworks:tube_1'},
+        {'technic:stainless_steel_ingot', 'technic:mv_cable',       'technic:stainless_steel_ingot'},
```
numberZero: Recipe is unreasonable

technic/machines/MV/freezer.lua
```diff
@@ -0,0 +1,12 @@
...
+    }
+})
+
+technic.register_freezer({tier = "MV", demand = {800, 600, 400}, speed = 2, upgrade = 1, tube = 1})
```
numberZero: Demand is small

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
+
+local S = technic.getter
+
+technic.register_recipe_type("freezing", { description = S("Freezing") })
+
+function technic.register_freezer_recipe(data)
+    data.time = data.time or 2
```
numberZero: Default time? That’s weird.
And it is way too short, if that’s seconds. Freezing is slow.

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
...
+    technic.register_recipe("freezing", data)
+end
+
+local recipes = {
+    {"bucket:bucket_water", { "default:snowblock", "bucket:bucket_empty" } },
+    {"bucket:bucket_river_water", { "default:snowblock", "bucket:bucket_empty" } },
+    {"default:dirt", "default:dirt_with_snow" }
```
numberZero: :+1:

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
...
+}
+
+for _, data in pairs(recipes) do
+    technic.register_freezer_recipe({input = {data[1]}, output = data[2]})
```
numberZero: Repeating `technic.register_freezer_recipe` would be okay IMO

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
...
+
+technic.register_recipe_type("freezing", { description = S("Freezing") })
+
+function technic.register_freezer_recipe(data)
+    data.time = data.time or 2
+    technic.register_recipe("freezing", data)
+end
+
+local recipes = {
+    {"bucket:bucket_water", { "default:snowblock", "bucket:bucket_empty" } },
```
numberZero: no-no, ice block, please.

numberZero: > That's right, but we have no default recipe to create snow.

Then add it, Grinding seems appropriate.

technic/machines/MV/freezer.lua
upsala: Replaced recipe with a new one

technic/machines/MV/freezer.lua
upsala: I don't think so, compressor has the same demand.

technic/machines/register/freezer_recipes.lua
upsala: Changed to 5s

technic/machines/register/freezer_recipes.lua
upsala: Done

upsala:
I've added 32px images, but I don't know, how to test them :)

technic/machines/register/freezer_recipes.lua
DS-Minetest: But to `dirt_with_snow`? Normally players don't gain this. Permafrost would make more sense, but ti's too new.

technic/machines/MV/freezer.lua
DS-Minetest: Lead could be used instead of stainless steel (or mix). I think, it insulates better than stainless steel. Please correct me, I might be wrong.

technic/machines/register/freezer_recipes.lua
upsala: I doesn't know permafrost. I've added dirt_with_snow only for completeness, maybe a mod uses it...

technic/machines/MV/freezer.lua
upsala: I've used stainless steel, because some freezers look like stainless steel...

technic/machines/register/freezer_recipes.lua
DS-Minetest: Permafrost was added to minetest_game in 5.0-dev: https://github.com/minetest/minetest_game/blob/21f5eaade052684d2694f6584496f8b5dc48d4ea/mods/default/nodes.lua#L499-L525

numberZero:
Sorry for long wait, was busy IRL. The PR looks okay. Now you need to find someone with push power... @SmallJoker ?

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
+
+local S = technic.getter
+
+technic.register_recipe_type("freezing", { description = S("Freezing") })
+
+function technic.register_freezer_recipe(data)
```
SmallJoker: Please extend https://github.com/minetest-mods/technic/blob/master/technic/doc/api.md#specific-machines accordingly.
`data` format validation (`assert` for example) would be nice to have as well.

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
...
+    data.time = data.time or 5
+    technic.register_recipe("freezing", data)
+end
+
+local recipes = {
+    {"bucket:bucket_water", { "default:ice", "bucket:bucket_empty" } },
```
SmallJoker: Please use tabs to indent.

technic/machines/register/freezer_recipes.lua
```diff
@@ -0,0 +1,20 @@
...
+    {"bucket:bucket_river_water", { "default:ice", "bucket:bucket_empty" } },
+    {"default:dirt", "default:dirt_with_snow" }
```
SmallJoker: The lava bucket could also be added to this list -> produces obsidian.

technic/machines/MV/freezer.lua
SmallJoker: Is it intended to be this fast? I think freezing should take much longer.

technic/machines/register/freezer_recipes.lua
upsala: This page needs much more updates and should be done in an seperate patch

tanmayameher:
Hell yeah. :+1: :100: 
Always wanted to make some cool ice easily to make an ice castle or something or may be an igloo :mountain_snow:

PR #466, opened 2018-12-03
fix #465

### LINKED ISSUE #465: Quarry facedir issue (server freeze)
A player of mine had the brilliant idea to turn his quarry so that it faces the bottom.
That caused a pretty weird server freeze (no errors or log)
Should be reproducible in singleplayer/standalone

thomasrudin:
Checks if facedir is valid. if the face points up or down the depth-check would run endless and freeze the server

### DIFF
--- a/technic/machines/HV/quarry.lua
+++ b/technic/machines/HV/quarry.lua
@@ -112,7 +112,7 @@ local function quarry_run(pos, node)
 
 	if meta:get_int("enabled") and meta:get_int("HV_EU_input") >= quarry_demand and meta:get_int("purge_on") == 0 then
 		local pdir = minetest.facedir_to_dir(node.param2)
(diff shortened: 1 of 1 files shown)


technic/machines/HV/quarry.lua
```diff
@@ -112,6 +112,11 @@ local function quarry_run(pos, node)
 
 	if meta:get_int("enabled") and meta:get_int("HV_EU_input") >= quarry_demand and meta:get_int("purge_on") == 0 then
 		local pdir = minetest.facedir_to_dir(node.param2)
+		if pdir.y > 0 or pdir.y < 0 then
```
DS-Minetest:
```suggestion
		if pdir.y ~= 0 then
```

Issue #472, opened 2018-12-23
Android won't change mining drill mode


Xiong:
The Android port is funny in several ways. Sneak is on onscreen overlay button. drill.lua checks keys.sneak but doesn't see the Android user. 

I've patched my own to change modes on_place (which on Android is double-tap, not right-click). This works well since I'm not in the habit of opening chests with my drill. 

Will submit PR.

Xiong:
Okay so that's a patch. Works on an existing drill (previously hacked to mode 5) and a newly-crafted drill. Should not interfere with existing behaviour for desktop users... but someone should check.

DustyBagel:
As far as I can tell, the mk2 and mk3 mining drills are able to switch modes just fine on desktop.

PR #473, opened 2018-12-23
Change drill mode on_place (double-tap on Android).

Xiong:
https://github.com/minetest-mods/technic/issues/472


DS-Minetest:
Fix the indention!

I don't like this change. It will also influence pc players.

Xiong:
I don't want to fool with indentation; I'm agnostic. I won't fight such crusade. 

If you don't like it, don't take it. Let Android burn.

SmallJoker:
> I won't fight such crusade.

Adaption welcome.

PR #482, opened 2019-01-23
Fix trailing whitespace

Hawk777:
I don’t know if you care about this, but I find trailing whitespace in files a little annoying. If you don’t care, please feel free to throw this PR away; it was just easier to submit the PR (knowing it might be dropped) than to go back and forth checking whether you wanted it or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,6 @@ FAQ
 License
 -------
 
-Unless otherwise stated, all components of this modpack are licensed under the 
+Unless otherwise stated, all components of this modpack are licensed under the
 LGPL, V2 or later.  See also the individual mod folders for their
 secondary/alternate licenses, if any.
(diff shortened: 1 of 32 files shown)


DS-Minetest:
I think, at markdown files, the spaces have a meaning, something with new lines.
Compare [yours](https://github.com/Hawk777/technic/blob/whitespace/manual.md#ore) with [the original](https://github.com/minetest-mods/technic/blob/master/manual.md#ore).

Hawk777:
You’re right, one of the lines did change appearance due to the removal of a hard linebreak. It appears that a *double* trailing space causes that, while a *single* trailing space doesn’t do anything. That said, the new format (without the linebreak) is in keeping with the other paragraphs in that area (about iron and copper) whose “It is a common metal” sentences are both *not* linebreak-separated from the preceding.

PR #483, opened 2019-01-23
Make power meter work on LV battery box

### LINKED ISSUE #440: LV battery box formspec charge meter doesn’t work
All battery boxes have a charge meter in their formspecs on the left side, labelled “Power level”. For the MV and HV battery boxes, the meter works fine. For the LV battery boxes, the meter always reads empty, regardless of the charge level of the box. The hover text and node texture show the proper charge level, and the battery box properly stores and release energy; only the meter is broken.

This happened on 9b40d02fbd9e8c39c5c64cc7de89a7fe45382f04.

Hawk777:
Fixes #440.

### DIFF
--- a/technic/machines/register/battery_box.lua
+++ b/technic/machines/register/battery_box.lua
@@ -129,11 +129,11 @@ local tube = {
 }
 
 local function add_on_off_buttons(meta, ltier, charge_percent)
-	local formspec = ""
-	if ltier == "mv" or ltier == "hv" then
-		formspec = "image[1,1;1,2;technic_power_meter_bg.png"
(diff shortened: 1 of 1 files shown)


DS-Minetest:
Looks good.
If this is tested and it works: :+1:

Hawk777:
Well, I tested it :grin: but someone else probably should.

DS-Minetest:
Btw. I've found the commit that introduced this (just to be sure there was no reason for this): https://github.com/minetest-mods/technic/commit/7244f86ee59f31a5eef9088c80436fb1edf1e38a

PR #498, opened 2019-05-27
add tubelib support to technic chests

obl3pplifp:
does what it says on the box

### DIFF
--- a/technic_chests/register.lua
+++ b/technic_chests/register.lua
@@ -2,6 +2,7 @@ local S = rawget(_G, "intllib") and intllib.Getter() or function(s) return s end
 
 local pipeworks = rawget(_G, "pipeworks")
 local fs_helpers = rawget(_G, "fs_helpers")
+local tubelib_exists = minetest.global_exists("tubelib")
 
 local allow_label = ""
 local tube_entry = ""
@@ -366,7 +367,7 @@ function technic.chests:register(name, data)
 	local nn = "technic:"..name:lower()..(data.locked and "_locked" or "").."_chest"
 	minetest.register_node(":"..nn, def)
 
-	if minetest.global_exists("tubelib") then
+	if tubelib_exists then
 		tubelib.register_node(nn, {}, _TUBELIB_CALLBACKS)
 	end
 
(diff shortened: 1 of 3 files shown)


technic_chests/register.lua
```diff
@@ -330,12 +330,48 @@ function technic.chests:definition(name, data)
 	return def
 end
 
+local _TUBELIB_CALLBACKS = {
+	on_pull_item = function(pos, side, player_name)
+		local inv = minetest.get_meta(pos):get_inventory()
```
SmallJoker: Move this inventory getter below the condition `is_protected`. It does not need to be executed in every case.

technic_chests/register.lua
```diff
@@ -330,12 +330,48 @@ function technic.chests:definition(name, data)
...
+	end,
+	on_unpull_item = function(pos, side, item, player_name)
+		local inv = minetest.get_meta(pos):get_inventory()
+		if inv:room_for_item("main", item) then
+			inv:add_item("main", item)
+			return true
+		end
+		return false
+	end,
+	on_recv_message = function(pos, topic, payload)
```
SmallJoker: According to https://github.com/joe7575/techpack/blob/master/tubelib/command.lua#L316 this can be left out if unused.

technic_chests/register.lua
```diff
@@ -330,12 +330,48 @@ function technic.chests:definition(name, data)
...
+	end
+}
+
 function technic.chests:register(name, data)
 	local def = technic.chests:definition(name, data)
 
 	local nn = "technic:"..name:lower()..(data.locked and "_locked" or "").."_chest"
 	minetest.register_node(":"..nn, def)
 
+	if minetest.get_modpath("tubelib") then
```
SmallJoker: Please make move this call into a local variable outside of `:register`.

technic_chests/register.lua
```diff
@@ -353,8 +389,10 @@ function technic.chests:register(name, data)
 			colordef.groups = self.groups_noinv
 			colordef.tiles = { def.tiles[1], def.tiles[2], def.tiles[3], def.tiles[4], def.tiles[5], mk_front("technic_chest_overlay"..postfix..".png") }
 			minetest.register_node(":"..nn..postfix, colordef)
+			if minetest.get_modpath("tubelib") then
```
SmallJoker: Use the local variable here. No need to call this function 15 times per node.

obl3pplifp:
Thanks for the feedback. I've made the changes you requested; as far as I can tell, checking for the tubelib modpath is superfluous and if you're worried about performance, just checking to see if the global exists should suffice. I readily admit I'm a neophyte and might be missing something though.

obl3pplifp:
I've made the requested change.

PR #500, opened 2019-06-15
Fix warnings about interlaced PNGs

cheapie:
There were three textures that were saved as interlaced PNGs. This is not necessary for MT and causes a warning ("Irrlicht: PNG warning: Interlace handling should be turned on when using png_read_image") to be emitted by the client. This commit replaces them with normal PNGs with the same content.

Unfortunately they did end up slightly larger than before - it would likely be possible to reduce this with some sort of optimization, but the total size for these three is still only about 7KB.


SmallJoker:
To minimize them, you could try to run pngcrush or `optipng -strip all` on them.

cheapie:
I'm pretty sure I tried running pngcrush on them already... if someone wants to try to optimize them better, go ahead, otherwise I'm just going to maintain that 7KB is not a big deal.

SmallJoker:
Replaced by #501

Issue #503, opened 2019-07-09
Protector block removal using two constructors


coil0:
On servers using the protector mod with a radius less than 8, it's possible to remove any protector block using two constructors and one blinky plant. First, placing a Mk3 constructor 8 blocks away from the protector (outside its radius), facing the protector. Then placing a blinky plant in the constructor's third slot and a second constructor in the fourth slot. When the first constructor is activated, it places the blinky plant and the second constructor inside the protected area. The blinky plant activates the second constructor, which removes the protector block.

If there are many protector blocks, someone sufficiently dedicated removes all of them starting from the border and removing one protector at a time.

The constructors do not have an owner, so they can hardly check for protection at the moment. Hard-coding names of protector blocks still allows removing other things in protected areas.

DS-Minetest:
Related: https://github.com/minetest-mods/mesecons/issues/281

PR #505, opened 2019-07-17
Reject constructor in constructor slot inventories

coil0:
This prevents using cascading constructors to get past protection, fixing #503.


DS-Minetest:
This might break builds, maybe.

coil0:
Tubes aren't able to add items to constructors and constructors already in another constructor's inventory remain there, so I don't think this can break builds.

PR #511, opened 2019-08-26
Add cool_trees support to the chainsaw

coil0:
This adds the nodes from the cool_trees modpack.


SmallJoker:
That's quite repetitive. How about making it dynamic so that adding more mods will be even easier?
```Lua
local nodes = {
	-- "node:name", is_leaf
	{ "bamboo:trunk",  false },
	{ "bamboo:leaves", true },
	{ "birch:trunk",   false },
	{ "birch:leaves",  true },
	-- ...
}

for i, d in ipairs(nodes) do
	if not d[2] or chainsaw_leaves then
		local mod_name = d[1]:match("^([^:]+)")
		if minetest.get_modpath(mod_name) then
			timber_nodenames[d[1]] = true
		end
	end
end
```

coil0:
We can make it still easier by not checking if the mod is installed, which seems unnecessary. The first part of the node name doesn't always match the mod that registers the node anyway.

PR #521, opened 2019-12-19
[WIP] Allowing tube.can_insert to be overridden for alloy furnaces

### LINKED ISSUE #520: Feature request: Alloy Furnace should refuse a second stack of identical items
No alloy recipe needs two identical components so that would make automation a lot easier.

Trying to figure out how it's possible to do, please add details that can help.

int-ua:
To reject second stacks of identical items.

Fixes #520

Credit to coil and SwissalpS for helping


int-ua:
Doesn't work as intended yet. Will have to compare counts as `inv_stack:item_fits` doesn't really work when stacks are allowed to be split.

PR #523, opened 2019-12-19
Avoiding alloy furnaces having identical items in both slots

### LINKED ISSUE #520: Feature request: Alloy Furnace should refuse a second stack of identical items

int-ua:
New version made with SwissalpS, supersedes #521 
Fixes #520

### DIFF
--- a/technic/doc/api.md
+++ b/technic/doc/api.md
@@ -64,6 +64,10 @@ Registration functions
 ### Specific machines
 * `technic.register_solar_array(data)`
 	* data is a table
+* `technic.can_insert_unique_stack(pos, node, stack, direction)`
+* `technic.insert_object_unique_stack(pos, node, stack, direction)`
+	* Functions for the parameters `can_insert` and `insert_object` to avoid
+	  filling multiple inventory slots with same type of item.
 
 Used itemdef fields
 -------------------
--- a/technic/machines/register/alloy_furnace.lua
+++ b/technic/machines/register/alloy_furnace.lua
@@ -1,10 +1,50 @@
 
 local S = technic.getter
 
+function technic.insert_object_unique_stack(pos, node, incoming_stack, direction)
+	local meta = minetest.get_meta(pos)
+	local inv = meta:get_inventory()
+	local incoming_name = incoming_stack:get_name()
+	local stack_index = nil
+	for inv_index, inv_stack in pairs(inv:get_list("src")) do
+		if inv_stack:get_name() == incoming_name then
+			stack_index = inv_index
+			break
+		end
+	end
+	if stack_index == nil then
+		return inv:add_item("src", incoming_stack)
+	end
+	local present_stack = inv:get_stack("src", stack_index)
+	local leftover = present_stack:add_item(incoming_stack)
+	inv:set_stack("src", stack_index, present_stack)
+	return leftover
+end
+
+function technic.can_insert_unique_stack(pos, node, incoming_stack, direction)
+	local meta = minetest.get_meta(pos)
+	local inv = meta:get_inventory()
+	local incoming_name = incoming_stack:get_name()
+	if meta:get_int("splitstacks") == 0 then
+		-- avoid looping second time with inv:contains_item("src", incoming_stack)
+		for _, inv_stack in pairs(inv:get_list("src")) do
+			if inv_stack:get_name() == incoming_name then
+				return inv_stack:item_fits(incoming_stack)
+			end
+		end
+	end
+
+	return technic.default_can_insert(pos, node, incoming_stack, direction)
+end
+
 function technic.register_alloy_furnace(data)
 	data.typename = "alloy"
 	data.machine_name = "alloy_furnace"
 	data.machine_desc = S("%s Alloy Furnace")
+
+	data.insert_object = technic.insert_object_unique_stack
+	data.can_insert = technic.can_insert_unique_stack
+
 	technic.register_base_machine(data)
 end
 
--- a/technic/machines/register/machine_base.lua
+++ b/technic/machines/register/machine_base.lua
@@ -4,22 +4,26 @@ local S = technic.getter
 local fs_helpers = pipeworks.fs_helpers
 local tube_entry = "^pipeworks_tube_connection_metallic.png"
 
-local tube = {
-	insert_object = function(pos, node, stack, direction)
-		local meta = minetest.get_meta(pos)
-		local inv = meta:get_inventory()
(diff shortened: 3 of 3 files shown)


technic/machines/register/alloy_furnace.lua
```diff
@@ -1,10 +1,58 @@
...
+end
+
+function technic.alloy_furnace_can_insert(pos, node, stack, direction)
+	local meta = minetest.get_meta(pos)
+	local inv = meta:get_inventory()
+	if meta:get_int("splitstacks") == 0 then
+		-- reject second stack of item that is already present
+		local incoming_name = stack:get_name()
+		for _, inv_stack in pairs(inv:get_list("src")) do
+			if not inv_stack:is_empty() and inv_stack:get_name() == incoming_name then
```
SmallJoker: 1) Can be shortened to `if inv_stack:get_name() == incoming_name then`
2) Why not use `inv:contains_item()` instead of this loop?
3) How does this even work? I think you're supposed to fill up the source list if the item stacks match.

technic/machines/register/alloy_furnace.lua
```diff
@@ -1,10 +1,58 @@
...
+	for _, inv_stack in pairs(inv:get_list("src")) do
+		local count = inv_stack:get_count()
+		if inv_stack:get_name() == incoming_name then
+			count_existing_of_incoming_type = count_existing_of_incoming_type + count
+			break
+		end
+	end
+	if 0 == count_existing_of_incoming_type then return inv:add_item("src", stack) end
+	local stack_max = stack:get_stack_max()
+	local overflow = count_existing_of_incoming_type + count_incoming - stack_max
```
SmallJoker: This entire overflow handling is already done by `inv:add_item()` like so:
```Lua
local leftover = inv:add_item("src", stack_to_add)
return leftover
```
Or alternatively for a specific stack index:
```Lua
local list = inv:get_list("src")
local leftover = list[stack_index]:add_item(stack_to_add)
inv:set_list("src", list)
return leftover
```

technic/machines/register/alloy_furnace.lua
int-ua: 1 and 2. Thank you, done.
3. `if meta:get_int("splitstacks") == 0`. I would like to not have this option at all in all technic machines but it's on by default. Is there a use-case for it?

technic/machines/register/alloy_furnace.lua
int-ua: Preliminary testing showed that this add_item overflow control fills second slot in "src" which is exactly the behavior we are trying to avoid. But working with single stack worked it seems, publishing new code. The delay of 12 days was cause by me not receiving a notification about these comments for some reason.

technic/machines/register/machine_base.lua
```diff
@@ -4,22 +4,26 @@ local S = technic.getter
 local fs_helpers = pipeworks.fs_helpers
 local tube_entry = "^pipeworks_tube_connection_metallic.png"
 
-local tube = {
+function technic.default_can_insert(pos, node, stack, direction)
```
SmallJoker: Please document this function in `technic/technic/api.md`

technic/machines/register/machine_base.lua
```diff
@@ -4,22 +4,26 @@ local S = technic.getter
...
+	local inv = meta:get_inventory()
+	if meta:get_int("splitstacks") == 1 then
+		stack = stack:peek_item(1)
+	end
+	return inv:room_for_item("src", stack)
+end
+
+function technic.new_default_tube()
+	return {
 	insert_object = function(pos, node, stack, direction)
```
SmallJoker: Missing indent for the code block until `}`. Whitespace changes can easily be ignored in the review process.

technic/machines/register/alloy_furnace.lua
```diff
@@ -1,10 +1,48 @@
...
+	local stack_index = nil
+	for checking_index, inv_stack in pairs(inv:get_list("src")) do
+		if inv_stack:get_name() == incoming_name then
+			stack_index = checking_index
+			break
+		end
+	end
+	if stack_index == nil then
+		return inv:add_item("src", stack)
+	else
```
SmallJoker: Superfluous `else`. This can be replaced with `end` to reduce the indents below.

int-ua:
I'm going to publish new commits here but do not merge this PR, I'll squish commits before merging and will create a new one before final merge.

technic/machines/register/machine_base.lua
int-ua: Will that work? (see commit)

int-ua:
Is there anything else I can improve before merging?

SmallJoker:
@int-ua If you're happy with the functionality, I can give it a try again to check whether it works as expected. Code style review will happen in the same step.

int-ua:
Yes, please do it. I'm keeping [WIP] to avoid merging before squashing commits manually to save SwissalpS' and your contribution in `Co-authored-by`, ok?

SmallJoker:
Is it possible to add a partial ItemStack to the alloy furnace's input list?
The PR works so far, but the following scenario should be improved if possible:

  * Alloy furnace: [20x tin ingot] [Empty]
  * Tube: 99x tin ingot

The 99 tin ingots are rejected, even though they could fill up the slot.

int-ua:
>     * Alloy furnace: [20x tin ingot] [Empty]
>     * Tube: 99x tin ingot
> 
> The 99 tin ingots are rejected, even though they could fill up the slot.

What is the state of "Allow splitting" setting? AFAIU you had it on off and it's working as intended, not splitting incoming stack, no?

technic/doc/api.md
```diff
@@ -64,6 +64,10 @@ Registration functions
 ### Specific machines
 * `technic.register_solar_array(data)`
 	* data is a table
+* `technic.alloy_furnace_can_insert(pos, node, stack, direction)`
+* `technic.alloy_furnace_insert_object(pos, node, stack, direction)`
```
SmallJoker: I think a more generic name would fit better - not only limited to alloy furnace.
Maybe like this?

 * `technic.can_insert_unique_stack` and
 * `technic.insert_object_unique_stack`

SmallJoker:
@int-ua Sorry, I totally forgot about that setting. It indeed works as expected now. There's only a single comment left over, everything else is ready to merge.

SmallJoker:
1a45ad19

PR #527, opened 2020-01-23
Add river water can

leftshift:
I noticed you couldn't pick up river water with the water can and decided to add another can for river water.
Feel free to propose changes to the recipe and such


SmallJoker:
Thanks. This PR looks quite good so far. Three remarks:

1) Please use tabs to indent the code
2) Run `optipng` or `pngcrush` on `technic_river_water_can.png` (less server traffic)
3) Add `technic_river_water_can.png` to [the README file](https://github.com/minetest-mods/technic/blob/master/technic/README.md)

leftshift:
should be all done. The biggest problem with the texture was leaving some unnecessary export options enabled which made it unnecessarily huge

PR #528, opened 2020-01-25
technic_cnc/cnc.lua: fix access to non-existant stack parameter

thetaepsilon-gamedev:
This implementation of allow_metadata_inventory_move appears to have been written like the other allow_* functions in this file, but for this particular function stack is not actually a parameter passed, only the actual count.
Return that instead of trying to reference a non-existant variable and crashing.


SmallJoker:
Looks good. Will merge this fix soon.
FYI: The author email address is not valid. It won't be assigned to your account.

PR #532, opened 2020-02-28
Deal with lava cans wasting lava

### LINKED ISSUE #531: Lava cans can waste lava
If you have use a full or partially full lava can to place lava, when the pointed node is a lava source node, the lava will disappear from the lava can but will not be placed into the world

NOTE: This might also apply to the water cans, it probably does, however, I haven't tested that, and that isn't as important, as you can make infinite water sources

BobFred7:
Should fix #531 
There are a bunch of unrelated commits here because I made the foolish mistake of committing to the master branch of my fork when dealing with a separate issue, but I think my reverts prevent that from causing any actual problems


technic/tools/cans.lua
```diff
@@ -52,7 +52,7 @@ function technic.register_can(d)
 			if def.on_rightclick and user and not user:get_player_control().sneak then
 				return def.on_rightclick(pos, minetest.get_node(pos), user, itemstack, pointed_thing)
 			end
-			if not def.buildable_to then
+			if not def.buildable_to or minetest.get_node(pos).name == data.liquid_source_name then
 				pos = pointed_thing.above
 				def = minetest.registered_nodes[minetest.get_node(pos).name] or {}
 				if not def.buildable_to then return end
```
DS-Minetest: You should probably add the check here too.
Otherwise it might be possible to waste lava when swimming in lava.

BobFred7: You mean on line 58?

DS-Minetest: Yes, line 58 is where I made the code comment.
Just do the same as on line 55.

BobFred7:
Been over a week and a half... any problems with this? is it going to be merged?

PR #534, opened 2020-03-08
Fix crash when placing cable plates

coil0:
All coordinates can be equal if pointed_thing.above is identical to pointed_thing.under. This [happens](https://github.com/SwissalpS/replacer/blob/608c8c42ac3154f3b9393a85773eb01361cdd101/replacer.lua#L190) when placing a cable plate with replacer.


SmallJoker:
> num = xyz[s(tostring(v):sub(-2, -2)..k)]

This line gives me anxiety.

PR #529, opened 2020-02-09
Use MT5 builtin’s translation system on all the modpack

louisroyer:
If you are using an already created map to test this PR, you may need to break and replace some nodes (like chests) to see effects.

PR #535, opened 2020-03-10
Change from intllib to minetest translations and update Spanish

CarlosBarrazaE:


louisroyer:
#529 also adds more translation strings related to chest:register

CarlosBarrazaE:
I didn't find #529 before, I don't know how to add it, because it lacks strings to translate, like the names of the digiline channels example: @ 1_battery_box technic/machines/register/battery_box.lua 318

louisroyer:
This string is used with
```lua
meta:set_string("channel", ltier.."_battery_box"..minetest.pos_to_string(pos))
```
`channel` is used by digiline and acts as an unique identifier. I may be wrong, but I think this string is never displayed and is used for internal purpose only. Even if it were printed, it may break things with digiline because users would not be able to write correctly this identifier in digiline code (using mesecon's luacontroller).

If they are any other string I forgot, please let me know.

CarlosBarrazaE:
That string is displayed when the drum channel is edited, the player can change it at any time.
Screenshot_20200313_142307
Switching station also has a switching_station.lua 46

louisroyer:
This means if you translate this string, its actual (=internal) initial value will be `@technic)lv_battery_box(-32,3,65)`. If the player do not edit this, and try to create a program in luacontroller using the displayed identifier `caja_de_bateria_bv(-32,3,65)`, the program will not work until the channel is edited. Translating the default channel name without breaking this require server side translation, but minetest translation api provide client side translation.
The case of switching station is similar to this one.

CarlosBarrazaE:
you're right, it should be blank as the force field because it can't be translated and there are players who think they are untouchable values

PR #542, opened 2020-04-07
Protector redo mod support for technic chests

Sires0:
Probably there is a lot to improve but here is my first attempt. I added an extra "flavour" of chest: Protected chests. Those are accessible by members of the area when protector mod is being used. Its very useful for teams, when you don't want to make like 20 keys for each person and hope they can manage to remember which key is for which chest.

The code I used to make it compatible with protector was very generic and probably could be used to make it compatible with other protection mods.

### DIFF
--- a/technic_chests/common.lua
+++ b/technic_chests/common.lua
@@ -36,6 +36,15 @@ local function inv_change(pos, count, player)
 	return count
 end
 
+local function inv_change_protected(pos, count, player)
+	-- Skip check for pipeworks (fake player)
+	if minetest.is_player(player) and
+			minetest.is_protected(pos, player:get_player_name()) then
+		return 0
+	end
+	return count
+end
+
 function technic.chests.inv_move(pos, from_list, from_index, to_list, to_index, count, player)
 	return inv_change(pos, count, player)
 end
@@ -44,6 +53,16 @@ function technic.chests.inv_put(pos, listname, index, stack, player)
 end
 function technic.chests.inv_take(pos, listname, index, stack, player)
 	return inv_change(pos, stack:get_count(), player)
+end
+
+function technic.chests.inv_move_protected(pos, from_list, from_index, to_list, to_index, count, player)
+	return inv_change_protected(pos, count, player)
+end
+function technic.chests.inv_put_protected(pos, listname, index, stack, player)
+	return inv_change_protected(pos, stack:get_count(), player)
+end
+function technic.chests.inv_take_protected(pos, listname, index, stack, player)
+	return inv_change_protected(pos, stack:get_count(), player)
 end
 
 function technic.chests.on_inv_move(pos, from_list, from_index, to_list, to_index, count, player)
--- a/technic_chests/register.lua
+++ b/technic_chests/register.lua
@@ -172,7 +172,10 @@ local function get_receive_fields(name, data)
 		end
 		if data.color then
 			-- This sets the node
-			local nn = "technic:"..lname..(data.locked and "_locked" or "").."_chest"
+			-- local nn = "technic:"..lname..(data.locked and "_locked" or "").."_chest"
+            local nn = "technic:"..name:lower()..
+                       (data.locked and "_locked" or "")..
+                       (data.protected and "_protected" or "").."_chest"
 			check_color_buttons(pos, meta, nn, fields)
 		end
 		if fields["fs_helpers_cycling:0:splitstacks"]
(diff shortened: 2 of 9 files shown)


technic_chests/common.lua
```diff
@@ -36,6 +36,15 @@ local function inv_change(pos, count, player)
 	return count
 end
 
+local function inv_change_protected(pos, count, player)
+	-- Skip check for pipeworks (fake player)
+	if minetest.is_player(player) and
```
Desour, line 41: Isn't this very insecure?
Fake players normally have a way to get some player name.

Sires0: First, I copied this from the original code just a few lines up there, so if its insecure, so is the original code. Also I don't think it checks for player name: https://dev.minetest.net/minetest.is_player
It checks if the object is a real player or not. This is useful because pipeworks uses fake players for some machines, so if they are there they *are* in fact supposed to be able to alter the content around it. So by checking if its an actual player it skips those machines, keeping them functional.

auouymous: Interesting, that means locked technic chests in unprotected areas can be emptied using a transfer mod and the chest can be stolen once empty. The default locked chest doesn't allow either, unless the transfer mod has the protection_bypass priv for its fake player.

auouymous:
Thank you! I've wanted to write this very same patch for quite a while.

Instead of checking for the protector mod, just wrap with a setting to enable the chests. This allows it to work with any protection mod, including the areas mod. If the protector logo is compatible with technic's license, include it with the patch and give attribution, or just create a new image.

technic_chests/common.lua
auouymous: The locked technic chests are secure and can't be dug when empty. A pipeworks filter injector placed by another player will not take from a locked technic chest. An "unprotected" protected technic chest can be pulled from but can't when sitting at the edge of a protection. So technic chests and these new protected chests are secure.

auouymous:
This is a problem with locked chests as well, but neither show locked or protected on the formspec. The ones with configurable names do show it (unless renamed), but an iron chest simply says "Iron Chest" in upper left corner, for all three variants.

technic_chests/register.lua
```diff
@@ -240,13 +243,22 @@ function technic.chests:definition(name, data)
 			pipeworks.after_place(pos)
 		end
 		table.insert(front, "technic_"..lname.."_chest_lock_overlay.png")
+	elseif data.protected then
+		locked_after_place = function(pos, _)
+			local meta = minetest.get_meta(pos)
+			meta:set_string("infotext",
+					S("%s Protected Chest"):format(name))
```
SmallJoker, line 250: This won't work. Use `@1` and put the replacement text into `S()`. `string.format` will modify the English template, and clients cannot translate that. Alternatively, do it like MTG/dyes: https://github.com/minetest/minetest_game/blob/master/mods/dye/init.lua#L107

technic_chests/register.lua
```diff
@@ -364,7 +389,9 @@ local _TUBELIB_CALLBACKS = {
 function technic.chests:register(name, data)
 	local def = technic.chests:definition(name, data)
 
-	local nn = "technic:"..name:lower()..(data.locked and "_locked" or "").."_chest"
+	local nn = "technic:"..name:lower()..
+			   (data.locked and "_locked" or "")..
+			   (data.protected and "_protected" or "").."_chest"
```
SmallJoker, line 394: It would be a good idea to move this value into `data`, so that the register functions can re-use this name.

technic_chests/silver_chest.lua
```diff
@@ -44,5 +56,26 @@ technic.chests:register("Silver", {
 	infotext = true,
 	color = false,
 	locked = true,
+	protected = false,
```
SmallJoker, line 59: `protected = false,` is superfluous. This argument should only be specified inside the `protector` case below.
Also consider re-using the definition table to avoid copy/pasta errors:
```Lua
local def = {
	width = 12,
	height = 6,
	sort = true,
	autosort = true,
	infotext = true,
	color = false,
	locked = true,
}
technic.chests:register("Silver", def)

if minetest.get_modpath("protector") then
	-- craft registration

	def = table.copy(def)
	def.protected = true
	technic.chests:register("Silver", def)
end
```

auouymous:
> Please add `protector` in `mod.conf`.

It should be made to work with any protection mod, including the `areas` mod.

Panquesito7:
It is added to `depends.txt`, it should be added to `mod.conf` either.

auouymous:
I'm saying it shouldn't be in either `depends.txt` or `mod.conf`, and the code shouldn't check for the `protector` modpath. There should be a setting to manually enable or disable the protected chests to allow them to work with _any_ protection mod.

SmallJoker:
PR neglected and needs changes. Please open a new one that addresses the comments, or ask for re-opening this one here.

PR #551, opened 2020-05-06
Chainsaw update (bushes and moretrees fixed)

dokutan:
This fixes the chainsaw support for moretrees and adds the default bushes.


SmallJoker:
The more often I see those fix PRs the more I think about automating the registration.
It would be very helpful when all nodes with the `tree` or `leaves` group could be added so that this list becomes rather superfluous. Just thinking loud - the changes look good.

dokutan:
The chainsaw now checks if the node is either in the list or belongs to the ``tree`` or ``leaves`` group. I have not touched the list (yet) as it might be useful to keep the known nodes for compatibility. Also the list is obvoiusly still required for all nodes that should be sawed but don't belong to either group (fruit and other decoration).

dokutan:
The shortened version looks great. Because i got annoyed by floating snow during testing, i have added a call to check_for_falling. From your side, is there anything else that should be done before merging?

SmallJoker:
> From your side, is there anything else that should be done before merging?

No, it looks good. I'll wait a few days so that testers have the chance to test the PR before it's merged. Just in case.

PR #557, opened 2020-05-25
add external machine registration support

KaylebJay:
Closes https://github.com/minetest-mods/technic/issues/419
Replaces https://github.com/minetest-mods/technic/pull/420

Does exactly what #420 was supposed to.
To test: Register a machine from a completely different mod and make sure that everything works.

### DIFF
--- a/technic/machines/register/machine_base.lua
+++ b/technic/machines/register/machine_base.lua
@@ -4,8 +4,6 @@ local S = technic.getter
 local fs_helpers = pipeworks.fs_helpers
 local tube_entry = "^pipeworks_tube_connection_metallic.png"
 
-data.modname = data.modname or "technic"
-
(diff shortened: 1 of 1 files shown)


technic/machines/register/machine_base.lua
```diff
@@ -4,6 +4,8 @@ local S = technic.getter
 local fs_helpers = pipeworks.fs_helpers
 local tube_entry = "^pipeworks_tube_connection_metallic.png"
 
+data.modname = data.modname or "technic"
```
SmallJoker, line 7: Sorry what? How's this supposed to work?
Did you mean to move this line into `technic.register_base_machine`?

KaylebJay: Doh, indeed. Good spot.

PR #564, opened 2020-09-03
constructor protection check

auouymous:
Check if the player who placed constructor can access protected positions before placing or removing nodes. A mk3 constructor can remove up to 4 nodes into a protection, including chests that contain items. This only fixes the protection bypass, but the constructor should probably call can_dig() before removing nodes to avoid losing items inside inventories.

Existing constructors won't have an owner but will continue to work. Only newly placed constructors will perform the protection check.

### DIFF
--- a/technic/machines/other/constructor.lua
+++ b/technic/machines/other/constructor.lua
@@ -103,7 +103,9 @@ local function make_on(mark, length)
 			minetest.check_for_falling(pos)
 			for i = 1, length do
 				place_pos = vector.add(place_pos, dir)
-				if owner ~= "" and minetest.is_protected(place_pos, owner) then return end
+				if owner ~= "" and minetest.is_protected(place_pos, owner) then
+					return
+				end
 				local place_node = minetest.get_node(place_pos)
 				deploy_node(inv, "slot"..i, place_pos, place_node, node)
 			end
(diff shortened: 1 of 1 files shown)


SmallJoker:
Thanks for the PR. Per-player ownership could be confusing if it is not visualized - i.e. not contained in `"infotext"`. Hence I would prefer to keep things simple by checking only `is_protected` without any specific ownership.

auouymous:
Calling is_protected without an owner would prevent the constructor from working inside protections, even when placed by the owner of the protection. The pipeworks node breaker appears to also set an owner when placed and its lack of displayed ownership isn't a problem. But I could display the owner in infotext if you want.

SmallJoker:
Oh sorry, I somehow assumed this was for an additional node ownership (dig/place). It's fine like this.

SmallJoker (approved):
LGTM. Will merge in a few days.

auouymous:
Thank you for modifying the get_player_name() line with empty quotes that I copy and pasted, it pointed out a bypass. You can use a constructor to place a constructor that will have an empty owner and bypass protection. You can't put a constructor directly inside the constructor, but the constructor will break a constructor and put it in its inventory. New commit always gives the constructor an invalid owner and a failure in after_place_node uses an invalid owner.

SmallJoker (approved):
LGTM. Will merge in a few days.

technic/machines/other/constructor.lua
```diff
@@ -102,6 +103,9 @@ local function make_on(mark, length)
 			minetest.check_for_falling(pos)
 			for i = 1, length do
 				place_pos = vector.add(place_pos, dir)
+				if owner ~= "" and minetest.is_protected(place_pos, owner) then
+					return
```
Panquesito7, line 107: Shouldn't `minetest.record_protection_violation()` be added here?
```suggestion
					minetest.record_protection_violation()
					return true
```

SmallJoker: This is pointless to call when an automated setup attempts to perform such map actions. The owner might not be online, and protection mods cannot do anything there.

PR #569, opened 2020-09-11
Prevent protection check on MV machine close.

### LINKED ISSUE #538: Prevent protection check on MV machine close
The protector mod has an option to flip player around if not allowed on the protection. Checking for `fields.quit` before calling the `pipeworks.may_configure()` function prevents a protection check and a player flip when closing an MV machine with a chest upgrade. I don't need credit.

http://borg.qzx.com/files/technic-avoid-protection-check-on-machine-close.patch

auouymous:
Fixes #538.

### DIFF
--- a/technic/machines/register/machine_base.lua
+++ b/technic/machines/register/machine_base.lua
@@ -209,10 +209,10 @@ function technic.register_base_machine(data)
 		after_place_node = data.tube and pipeworks.after_place,
 		after_dig_node = technic.machine_after_dig_node,
 		on_receive_fields = function(pos, formname, fields, sender)
-			local node = minetest.get_node(pos)
 			if fields.quit then return end
 			if not pipeworks.may_configure(pos, sender) then return end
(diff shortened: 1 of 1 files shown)


technic/machines/register/machine_base.lua
```diff
@@ -210,7 +210,7 @@ function technic.register_base_machine(data)
 		after_dig_node = technic.machine_after_dig_node,
 		on_receive_fields = function(pos, formname, fields, sender)
 			local node = minetest.get_node(pos)
-			if not pipeworks.may_configure(pos, sender) then return end
+			if fields.quit or not pipeworks.may_configure(pos, sender) then return end
```
SmallJoker, line 213: Independent checks. I'd prefer to see them separate.
```suggestion
			if fields.quit then return end
			if not pipeworks.may_configure(pos, sender) then return end
```

Please also do this change for the `_active` node.

auouymous: Done.

PR #571, opened 2020-09-18
Update quarry.lua

yodabear1:
walking_light adds enhanced torches by adding virtual light sources.
let's tread them like air.


SmallJoker:
Thanks for your contribution.
This PR surely prevents that walking_light nodes aren't dug by quarries, but it does not fix the root of the problem. Generally all `drawtype == "airlike"` should be skipped from digging.
It's a pity that `walking_light` uses a 0-sized "glasslike" node rather than "airlike" directly. This mod from 2012 needs a serious rewrite in general.

Sorry, I don't think it makes sense to merge this PR in the current state.

PR #574, opened 2020-12-20
nuclear_reactor: add error messages on start failure

cedrichaase:
I built a nuclear reactor a few days ago - when trying to start it, I just simply got the (pretty much useless) message "Error" printed to the console.

I had to read the source to find out that I actually need to insert 6 pieces of uranium fuel into the reactor (as opposed to just one).

I was hoping to save anyone who wants to build nuclear reactors in the future the trouble of jumping through the same hoops, so here are some slightly more useful error messages.

### DIFF
--- a/technic/machines/HV/nuclear_reactor.lua
+++ b/technic/machines/HV/nuclear_reactor.lua
@@ -234,7 +234,7 @@ local function start_reactor(pos, meta)
 		end
 	end
 	-- Check that the reactor is complete and has the correct fuel
-	if correct_fuel_count ~= 6 or reactor_structure_badness(pos) ~= 0 then
+	if correct_fuel_count ~= 6 then
 		return msg_fuel_missing
 	end
 
--- a/technic/machines/HV/nuclear_reactor.lua
+++ b/technic/machines/HV/nuclear_reactor.lua
@@ -217,7 +217,7 @@ end
 
 
 local function start_reactor(pos, meta)
-    msg_fuel_missing = "Error: You need to insert 6 pieces of Uranium Fuel."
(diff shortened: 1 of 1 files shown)


technic/machines/HV/nuclear_reactor.lua
```diff
@@ -233,16 +235,22 @@ local function start_reactor(pos, meta)
 	end
 	-- Check that the reactor is complete and has the correct fuel
 	if correct_fuel_count ~= 6 or reactor_structure_badness(pos) ~= 0 then
-		return false
+		return msg_fuel_missing
+	end
+
+	if reactor_structure_badness(pos) ~= 0 then
```
SmallJoker, line 241: This is already covered by the check above.

technic/machines/HV/nuclear_reactor.lua
```diff
@@ -217,12 +217,14 @@ end
 
 
 local function start_reactor(pos, meta)
+	local msg_fuel_missing = "Error: You need to insert 6 pieces of Uranium Fuel."
```
SmallJoker, line 220: `Insufficient Uranium fuel` would be future-proof in case the cost is balanced somewhen.

Trivial, decision up to you.

cedrichaase: Good point, I guess I should put the magic number 6 into a variable and use it both for the algorithm and message

technic/machines/HV/nuclear_reactor.lua
cedrichaase: Thanks. I removed the check above so that we can fall through to this check here instead and return a more useful message.

SmallJoker:
:+1: Looks good. Will merge it in a few days unless there are objections.

PR #575, opened 2020-12-20
centrifuge: reduce power by half

cedrichaase:
I guess the contents of this PR are pretty much self-explanatory, so I'll just share a little bit about why I feel this is more adequate :)

In a typical minetest/technic-game, enriching uranium seems to be the most common use case for centrifuges. For a fully automated uranium processing pipeline (0,0 .. 3,5% fissile uranium dust), a total of 34 centrifuges are required. This is, imo, the only way of enriching uranium to 3,5% without going insane. At this point, as a player, I have expended 136 diamonds just for building centrifuges (4 each) and a ton of effort for actually crafting and hooking up the centrifuges with sorting tubes and the likes.
I feel that, at this point, the game should reward players for putting in the effort into building a machine like this.

The resulting machine, however, then has a typical power consumption of ~100kEU, which is roughly equivalent to the output of a nuclear reactor. In order to satisfy this consumption, the player would need to construct a power plant (using solar panels, water mills etc.) that has the same output as the nuclear reactor they were trying to construct in the first place, which leads the entire project ad absurdum.

I feel that by drastically reducing centrifuge power consumption, progression towards building a nuclear reactor would make a lot more sense. It would still require players to build a sizeable power plant before moving on towards a nuclear reactor, but not one with equivalent power output.

### DIFF
--- a/technic/machines/MV/centrifuge.lua
+++ b/technic/machines/MV/centrifuge.lua
@@ -9,7 +9,7 @@ minetest.register_craft({
 
 technic.register_centrifuge({
 	tier = "MV",
-	demand = { 8000, 7000, 6000 },
+	demand = { 4000, 3500, 3000 },
 	speed = 2,
 	upgrade = 1,
 	tube = 1,


auouymous:
> 34 centrifuges are required

Only 3 are actually required, your choice of 34 is for speed when processing very large quantities.

A chest with 3 stackwise injectors each feeding a single centrifuge which auto-ejects (control logic unit + battery) back to the chest. Add 2 of each dust in the injectors, staggering the purity: 0.1 in first, 0.2 in second, 0.3 in third, 0.4 in first, and so on. This setup is very compact and uses a fraction of the power.

OgelGames:
From https://github.com/minetest-mods/technic/blob/master/manual.md:
> A centrifuge cascade will consume quite a lot of energy. It is worth putting a battery upgrade in each centrifuge. (Only one can be accommodated, because a control logic unit upgrade is also required for tube operation.) An MV centrifuge, the only type presently available, draws 7 kEU/s in this state, and takes 5 s for each uranium centrifuging operation. It thus takes 35 kEU per operation, and the cascade requires 17.15 MEU to produce each unit of enriched uranium. It takes five units of enriched uranium to make each fuel rod, and six rods to fuel a reactor, so **the enrichment cascade requires 514.5 MEU to process a full set of reactor fuel. This is about 0.85% of the 6.048 GEU that the reactor will generate from that fuel.**

cedrichaase:
Thanks to both of you for your input. I'm getting the impression that the problem I've come across might be better remedied by more extensive and visual documentation, as opposed to just textual descriptions. I'll look through issues concerning this and might open one (or maybe even a PR to the Wiki) depending on what I find. :slightly_smiling_face:

PR #577, opened 2021-01-08
Supply converter demands only needed power

### LINKED ISSUE #576: Converters always consume maximal power
Voltage converters always consume their maximal power, even if the machines it powers consume less.

In real world, a voltage transformer consumes as much as much energy as we demand to it (excluding heat and noise).

A minimal consumption may be added (for controller electronics and noise) but this always-maximal consumption is annoying.

ZettaScript:
Partially fixes #576.

Supply converter demands only the needed power to satisfy the demand on the output network. It can still be too much if the output network has other power sources, because supply is not taken in account.

To test it: place a powerful generator at the converter's input and a little machine at it's output. e.g. if the machine consumes 300 EU, the converter will demand 400 EU instead of the maximal power. This enables using the remaining power on the input network.

This is the first time I write Lua, and my first contribution to a Minetest mod so I'm not sure it's ok, but it works on my server.

Edit: it causes a problem: battery boxes are not charged on the output network. To fix that, I think battery boxes should use the same demand/supply system as the other machines. A not fully charged battery box demands a fixed power, and a not empty battery box supplies a fixed power.

### DIFF
--- a/technic/machines/supply_converter.lua
+++ b/technic/machines/supply_converter.lua
@@ -133,7 +133,6 @@ local run = function(pos, node, run_stage)
 		enabled = enabled == "1"
 	end
 	enabled = enabled and (meta:get_int("mesecon_mode") == 0 or meta:get_int("mesecon_effect") ~= 0)
-	local demand = enabled and meta:get_int("power") or 0
 
 	local pos_up        = {x=pos.x, y=pos.y+1, z=pos.z}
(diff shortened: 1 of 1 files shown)


ZettaScript:
I think of a simpler fix to allow supply converters to charge batteries: the supply converter just counts the number of batteries on the output network, then add this number times the battery's power to it's own demand.

SmallJoker (approved):
Works. Thank you.

Issue #474, opened 2018-12-24
Grinder required to construct a Grinder


cyisfor:
* You can't use a grinder without a switching station. 
* A switching station requires a LV transformer to construct. 
* A LV transformer requires rubber fiber. 
* Rubber fiber requires coal dust. 
* Coal dust can only be produced with a grinder.

Constructing an LV transformer has required rubber since March of 2016, so... nobody has ever tried to use technic outside of creative mode for the past 2 or 3 years? Because unless I'm missing something, it's completely unusable.

VanessaE:
Maybe your copy of Technic is modified somehow, but you're missing one detail: it relies on mesecons fiber, not rubber fiber.  So this crafting chain works:

7 plastic -> empty spools
1 copper ingot + two empty spools -> spools of copper wire
4 spools of copper wire + 4 steel ingot -> copper coil (you get the empty spools back)
1 sapling or 1 raw latex -> cooking -> glue
glue -> cooking -> Mesecons fiber

2 copper coil + 5 steel ingot + 2 Mesecons fiber -> LV transformer

steel ingot -> cooking -> cast iron ingot
8 brick blocks -> fuel fired alloy furnace
2 copper ingot + 1 zinc ingot -> alloying -> brass ingots
1 brass ingot + 8 cast iron ingots -> machine casing

6 paper + 3 copper ingots -> LV cables

2 copper ingot + 1 LV transformer + 1 machine casing + 3 LV cable -> switching station

BobFred7:
Should we close this?

BobFred7:
Hold on: https://github.com/minetest-mods/technic/blob/master/technic/crafts.lua#L106
The transformer recipe (shortly after that line) uses an item that is a choice between mesecons fiber (what most technic installations use, and what VE thinks is correct) and technic rubber fiber (what cysifor is having a problem with). Cysifor's problem occurs whenever you are using technic without mesecons_materials.

TechDudie:
Maybe mesecons_materials should be a dependency.

SmallJoker:
#578 merged

PR #578, opened 2021-01-24
Fix issue #474 by adding a crafting recipe for  technic:rubber without the grinder

TechDudie:


SmallJoker:
Why? (general)

And why only for 0.4.x clients? `depends.txt` is ignored when `mod.conf` is found within the mod.

TechDudie:
So issue #474 can be closed. When I first downloaded technic and the dependencies, I couldn’t make a few things. I went here and then realized I needed mesecons_materials. This is annoying for people who are new to technic.

SmallJoker:
Technic is already quite dependency-heavy. If it's within your capabilities, I'd like it very much to have an alternative recipe for `technic:raw_latex 4` -> furnace cooking -> `technic:rubber 4` (or 2) if  `mesecons_materials` is not available.

TechDudie:
Hmmmm..... maybe a new solution can be used instead of a new dependency.....

TechDudie:
Perhaps you can craft 4 raw latex and 5 coal into 2 rubber things.....

TechDudie:
It’s “collecting” the loose bits of coal that come off of the lumps.

TechDudie:
Actually on second though, it crafts a rubber goo that can be smelted into rubber. I’ll add the changes now.

TechDudie:
Ok, changes made.

TechDudie:
Oops......

SmallJoker (approved):
LGTM. WIll merge in a few days if there are no objections.

TechDudie:
Ok

PR #581, opened 2021-02-12
granite, marble and marble bricks compat with underch mod

NathanielFreeman:
If [underch](https://content.minetest.net/packages/Hume2/underch/) exist already add marble, marble bricks and granite, technic add compat alias and not register that node.

alias allow retro compatibility :)

### DIFF
--- a/technic_worldgen/nodes.lua
+++ b/technic_worldgen/nodes.lua
@@ -46,6 +46,16 @@ minetest.register_node( ":technic:mineral_sulfur", {
 	drop = "technic:sulfur_lump",
 })
 
+if minetest.get_modpath("underch") then
+
+   minetest.register_alias("technic:granite", "underch:granite")
+
+   minetest.register_alias("technic:marble", "underch:marble")
+
+   minetest.register_alias("technic:marble_bricks", "underch:marble_brick")
+   
+else
+   
 minetest.register_node( ":technic:granite", {
 	description = S("Granite"),
 	tiles = { "technic_granite.png" },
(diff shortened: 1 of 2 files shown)


technic_worldgen/mod.conf
```diff
@@ -1,3 +1,3 @@
 name = technic_worldgen
 depends = default, basic_materials
-optional_depends = intllib, mg, doors, farming, glooptest, mesecons_doors, vessels
+optional_depends = intllib, mg, doors, farming, glooptest, mesecons_doors, vessels, underch
```
SmallJoker, line 3: So what about `depends.txt` for 0.4.x clients? As long that file exists within the modpack, it should be updated as well.

technic_worldgen/nodes.lua
```diff
@@ -46,6 +46,16 @@ minetest.register_node( ":technic:mineral_sulfur", {
...
 })
 
+if minetest.get_modpath("underch") then
+
+   minetest.register_alias("technic:granite", "underch:granite")
+
+   minetest.register_alias("technic:marble", "underch:marble")
+
+   minetest.register_alias("technic:marble_bricks", "underch:marble_brick")
+
```
SmallJoker, line 56: Please use tabulators to indent

technic_worldgen/nodes.lua
```diff
@@ -46,6 +46,16 @@ minetest.register_node( ":technic:mineral_sulfur", {
...
+   minetest.register_alias("technic:granite", "underch:granite")
+
+   minetest.register_alias("technic:marble", "underch:marble")
+
+   minetest.register_alias("technic:marble_bricks", "underch:marble_brick")
+   
+else
+   
 minetest.register_node( ":technic:granite", {
```
SmallJoker, line 59: else-case would be easier to read with an additional indent level. Whitespaces can be ignored during review, so that's not a problem.

Same for L194 until EOF

SmallJoker:
Please check your editor settings. It seems to mix spaces and tabs randomly.

PR #584, opened 2021-03-26
Add support for craftguide mod recipe registration.

sfence:
So, technic crafting recipes should be visible in [Crafting Guide](https://content.minetest.net/packages/jp/craftguide/) like they are visible in unified_inventory.


S-S-X:
Have to add craftguide to .luacheckrc read_globals and fix this:
```
    technic/machines/register/recipes.lua:77:11: shadowing definition of variable index on line 52
```

technic/machines/register/recipes.lua
```diff
@@ -59,6 +67,27 @@ local function register_recipe(typename, data)
...
+        result = result[1];
+      end
+      local items = "";
+      for index, input in pairs(data.input) do
+        if (items=="") then
+          items = items..input;
+        else
+          items = items..", "..input;
+        end
+      end
```
S-S-X, line 83: Use `_` instead of `index` as index value is not actually used at all in for loop.
Underscore has special meaning for linter: clears unused variable warning.
Also clears warnings about shadowing previous definition.
Also spaces -> tabs as to keep indentation similar how it is in other parts.

S-S-X (changes requested):
Requested small changes to clear linter tests, otherwise looks okay for me.

However I think it would be good to move both craftguide and unified_inventory stuff to separate file to keep technic specific registrations bit cleaner but IMO that can be done later too.
Maybe just add issue for that if not doing under this PR.

technic/machines/register/recipes.lua
sfence: Thanks for the quick reaction.
I have updated .luacheckrc, replaced spaces with tabs and the index variable is by _.

sfence:
I have added support for i3 inventory mod delivered from craftguide mod too.

technic/machines/register/recipes.lua
```diff
@@ -59,6 +71,36 @@ local function register_recipe(typename, data)
 			width = 0,
 		})
 	end
+	if (have_cg or have_i3) and technic.recipes[typename].output_size == 1 then
+		if (craftguide.register_craft or have_i3) then
```
SmallJoker, line 75: /recipes.lua:75: attempt to index global 'craftguide' (a nil value)

technic/machines/register/recipes.lua
```diff
@@ -59,6 +71,36 @@ local function register_recipe(typename, data)
...
+				result = result[1];
+			end
+			local items = "";
+			for _, input in pairs(data.input) do
+				if (items=="") then
+					items = items..input;
+				else
+					items = items..", "..input;
+				end
+			end
```
SmallJoker, line 87: `local items = table.concat(data.input, ", ")`

technic/machines/register/recipes.lua
```diff
@@ -59,6 +71,36 @@ local function register_recipe(typename, data)
...
+			if craftguide.register_craft then
```
SmallJoker, line 88: /recipes.lua:88: attempt to index global 'craftguide' (a nil value)

technic/machines/register/recipes.lua
```diff
@@ -13,6 +15,16 @@ function technic.register_recipe_type(typename, origdata)
 			height = 1,
 		})
 	end
+	if have_cg and craftguide.register_craft_type and data.output_size == 1 then
+		craftguide.register_craft_type(typename, {
+			description = data.description,
+		})
+	end
+	if have_i3 and i3.register_craft_type and data.output_size == 1 then
```
SmallJoker, line 23: I don't think checking for `register_craft_type` is necessary here. All (reasonable new?) versions have this API function.

2nd suggestion: Move the `data.output_size == 1` check one level up, and include all craft guide registrations inside to avoid duplication.

technic/depends.txt
```diff
@@ -12,3 +12,4 @@ intllib?
 unified_inventory?
 vector_extras?
 dye?
+craftguide?
```
SmallJoker, line 15: You might as well remove this file. technic now requires Minetest 5.0.0 anyway, hence this is superfluous.

sfence: The file has been removed by commit 7600b7e.

technic/machines/register/recipes.lua
sfence: My fault. I forgot to test it without craft guide mod enabled.

Should be fixed now.

technic/machines/register/recipes.lua
```diff
@@ -59,6 +73,29 @@ local function register_recipe(typename, data)
 			width = 0,
 		})
 	end
+	if (have_cg or have_i3) and technic.recipes[typename].output_size == 1 then
+		if (have_cg or have_i3) then
```
SmallJoker, line 77: Duplicated `(have_cg or have_i3)` check.

technic/machines/register/recipes.lua
```diff
@@ -59,6 +73,29 @@ local function register_recipe(typename, data)
...
+			local result = data.output;
+			if (type(result)=="table") then
+				result = result[1];
+			end
+			local items = table.concat(data.input, ", ");
```
SmallJoker, line 82: FYI: Lua does only require semicolons as long you're not doing multiple assignments per line (which you shouldn't do anyway).

This is not a change request.

technic/machines/register/recipes.lua
sfence: Fixed.

SmallJoker:
Thanks. Looks good. Will merge in a few days if there are no objections.

PR #593, opened 2021-10-27
Add granite brick node.

sfence:
Add granite brick building node.


technic_worldgen/nodes.lua
```diff
@@ -54,6 +54,14 @@ minetest.register_node( ":technic:granite", {
 	sounds = default.node_sound_stone_defaults(),
 })
 
+minetest.register_node( ":technic:granite_bricks", {
+	description = S("Granite Bricks"),
+	tiles = { "technic_granite_bricks.png" },
+	is_ground_content = true,
```
SmallJoker, line 60: No, this node is not part of the map generator.

sfence: It is only copied and changed version of marble_bricks.

https://github.com/minetest-mods/technic/blob/49d4105a2b7b782e49d68db0229b2a64542bf5b5/technic_worldgen/nodes.lua#L65-L71

So, should I move it and marble_bricks somewhere else or should I change is_ground_content to false for granite and marble bricks?

SmallJoker: I see that the same comment was raised in mt-mods/technic two years ago. To be entirely correct, both bricks would use `is_ground_content = false`. It only does make a difference when `/deleteblocks` is used or the mapgen limit setting increased - which could overwrite existing buildings close to map borders.

sfence: Ok. should be fixed in the following commit.

SmallJoker:
Ran `optipng -o7 -strip all`. Squashed and pushed as commit a296446.

PR #600, opened 2022-05-31
Add charge set/get xcallbacks

sfence:
Add technic_set_charge and technic_get_charge optional callbacks for tems/tools.

For the easier making of multi mod chargeable tools.

### DIFF
--- a/technic/doc/api.md
+++ b/technic/doc/api.md
@@ -144,6 +144,10 @@ Additional definition fields:
 	* Specifies the machine's node name to use when it's not connected connected to a network
 * `<itemdef>.technic_on_disable = function(pos, node) ...`
 	* This callback is run when the machine is no longer connected to a technic-powered network.
+* `<itemdef>.technic_get_charge = function(itemstack) ...`
+	* This optional callback will be used to get itemstack charge and max\_charge
+* `<itemdef>.technic_set_charge = function(itemstack, charge) ...`
+	* This optional callback will be used to set itemstack charge
 
 
 ## Node Metadata fields
--- a/technic/machines/register/battery_box.lua
+++ b/technic/machines/register/battery_box.lua
@@ -416,18 +416,29 @@ function technic.charge_tools(meta, batt_charge, charge_step)
 	end
 	local src_stack = inv:get_stack("src", 1)
 
-	local tool_name = src_stack:get_name()
-	if not technic.power_tools[tool_name] then
-		return batt_charge, false
-	end
-	-- Set meta data for the tool if it didn't do it itself
-	local src_meta = minetest.deserialize(src_stack:get_metadata()) or {}
-	if not src_meta.charge then
-		src_meta.charge = 0
+	-- get tool charge
+	local item_max_charge
+	local tool_charge
+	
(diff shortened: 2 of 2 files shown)


technic/machines/register/battery_box.lua
```diff
@@ -416,18 +416,29 @@ function technic.charge_tools(meta, batt_charge, charge_step)
...
+	local item_max_charge
+	local tool_charge
+
+	local src_def = src_stack:get_definition()
+	if src_def.technic_get_charge then
+		-- use get_charge callback
+		tool_charge, item_max_charge = src_def.technic_get_charge(src_stack)
+	else
+		-- check if is chargable
+		local tool_name = src_stack:get_name()
```
SmallJoker, line 429: What would you think of moving this code block to a local function? This way you could do the following unified call:

```Lua
local technic_get_charge = src_def.technic_get_charge or default_get_charge
local tool_charge, item_max_charge = technic_get_charge(src_stack)
```

and the same for the setter below. This will keep the parameters (1) and return values (2) perfectly in sync, which is especially helpful if new functionality (i.e. params) are added to the callback in the future.

technic/doc/api.md
```diff
@@ -144,6 +144,10 @@ Additional definition fields:
 	* Specifies the machine's node name to use when it's not connected connected to a network
 * `<itemdef>.technic_on_disable = function(pos, node) ...`
 	* This callback is run when the machine is no longer connected to a technic-powered network.
+* `<itemdef>.technic_get_charge = function(itemstack) ...`
+	* This optional callback will be used to get itemstack charge and max\_charge
```
SmallJoker, line 148: Needs return value documentation.

technic/machines/register/battery_box.lua
sfence: Should be fixed by [5c34070](https://github.com/minetest-mods/technic/pull/600/commits/5c34070bdce9f29a285ec3c213848fc219644a54)

technic/doc/api.md
sfence: Should be fixed by [5c34070](https://github.com/minetest-mods/technic/pull/600/commits/5c34070bdce9f29a285ec3c213848fc219644a54)

technic/machines/register/battery_box.lua
```diff
@@ -409,25 +409,48 @@ minetest.register_on_player_receive_fields(
...
+	return item_meta.charge, technic.power_tools[tool_name]
+end
+
+local function default_set_charge(itemstack, charge)
+	local tool_name = itemstack:get_name()
+	if technic.power_tools[tool_name] then
+		technic.set_RE_wear(itemstack, charge, technic.power_tools[tool_name])
+	end
+	local item_meta = {charge = charge}
+	itemstack:set_metadata(minetest.serialize(item_meta))
```
SmallJoker, line 432: This discards all fields other than `charge`. Is that intended?

SmallJoker:
Thank you. I noticed that the previous code did the same thing, hence removed the comment shortly afterwards. Nevertheless, a good addition.

PR #612, opened 2022-11-02
Add register_cable_tier API function

sfence:
Add register_cable_tier API function to allow user defined cables registration.

I need a special cable in "node" for my mod cleanroom, which I am working on.

### DIFF
--- a/technic/machines/register/cables.lua
+++ b/technic/machines/register/cables.lua
@@ -11,6 +11,10 @@ function technic.get_cable_tier(name)
 	return cable_tier[name]
 end
 
+function technic.register_cable_tier(name, tier)
+	cable_tier[name] = tier
+end
+
 local function check_connections(pos)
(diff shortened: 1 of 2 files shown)


technic/machines/register/cables.lua
```diff
@@ -11,6 +11,10 @@ function technic.get_cable_tier(name)
 	return cable_tier[name]
 end
 
+function technic.register_cable_tier(name, tier)
+	cable_tier[name] = tier
```
SmallJoker, line 15: This to enforce correct API usage early on to inform modders about potential issues.

```suggestion
	assert(technic.machines[tier], "Tier does not exist")
	assert(type(name) == "string", "Invalid node name")

	cable_tier[name] = tier
```

sfence:
Done and test. Is loadable and runnable.

Issue #601, opened 2022-06-08
Technic general TODO list


SmallJoker:
Help us to make Technic a better mod! But what to do?

- [ ] Incrementally switch to `.tr` client-side translations. Deprecate intllib.
    * Hints: `minetest.get_translator`, https://github.com/minetest-tools/update_translations
- [ ] Replace `itemstack:get_metadata()` with `itemstack:get_meta()`
    * Needs a compatibility function to migrate existing items, for example `technic.get_stack_meta_compat(itemstack)`
- [ ] General code maintenance.
    * De-duplicate code
    * Move repeated operations into variables or functions
    * Check for unused functions and purge them
- [ ] Textures: The furnace front texture does not match the sides. Needs color adjustment and/or new textures.

Each point will need a separate Pull Request (and commit). This is my personal list, but the tasks will not be done anytime soon, hence contributions are much appreciated!

OgelGames:
The first task already has a PR: https://github.com/minetest-mods/technic/pull/529

MidnightPhantom:
Hey Please add the recycler, that would turn your trash into basic materials

PR #628, opened 2024-01-08
Get rid of deprecated metadata

cx384:
This solves a part of #601

The deprecated metadata gets converted to a proper ItemStackMetaRef.
All keys stay the same except for:
- Cans that use `can_level` now, since they didn't store a serialized table in the metadata before.
- `charge` which is now `technic:charge`, since any item (also from other mods) may have a technic charge which can cause compatibility problems.

(For other keys e.g. `mode`, the key is special to the item and not universally used, which is not that problematic, I guess.)

There is now only one place left where metadata is used, the sorting function of techinc chests:
https://github.com/minetest-mods/technic/blob/5826c2feaa5aea2ade07af93cfec05952470c17b/technic_chests/register.lua#L149-L150
However, I'm unsure how this should be changed.
Getting the whole stack meta as a table and use the amount of keys or a serialized version to sort, may be an option, but I don't think it is justified performance-wise.
## To do

This PR is a Ready for Review.

(Maybe change the techinc chests sorting function.)

## How to test

1. Start a game with the old version and get some tools, charge them up, change their modes, fill up some cans.
2. Switch to the new version and see if the tools still work as expected.

I tested it, and didn't find any problem, but I'm not sure if I missed something.

## Compatibility note
Other mods which directly accessed the item metadata of technic items will not work anymore.

### DIFF
--- a/technic/legacy.lua
+++ b/technic/legacy.lua
@@ -55,12 +55,7 @@ function technic.get_stack_meta(itemstack)
 		if legacy_table then
 			local table = meta:to_table()
 			for k, v in pairs(legacy_table) do
-				local legacy_key = technic.legacy_meta_keys[k]
-				if legacy_key then
-					table.fields[legacy_key] = v
-				else
-					table.fields[k] = v
-				end
+				table.fields[technic.legacy_meta_keys[k] or k] = v
 			end
 			meta:from_table(table)
 		end
--- a/technic/legacy.lua
+++ b/technic/legacy.lua
@@ -39,3 +39,33 @@ for i = 0, 64 do
 	minetest.register_alias("technic:lv_cable"..i, "technic:lv_cable")
 end
 
+-- Converts legacy itemstack metadata to itemstack meta and returns the ItemStackMetaRef
+function technic.get_stack_meta_compat(itemstack)
+	local meta = itemstack:get_meta()
+	local metadata = meta:get("")
+	if metadata then
+		local metadata_table = minetest.deserialize(metadata)
+		if metadata_table then
+			local table = meta:to_table()
+			for k, v in pairs(metadata_table) do
+				table.fields[k] = v
+			end
+			meta:from_table(table)
+		end
+		meta:set_string("", "")
+	end
+	return meta
+end
+
+-- Same as technic.get_stack_meta_compat for cans.
+-- (Cans didn't store a serialized table in the metadata, but just a number.)
+function technic.get_stack_meta_compat_cans(itemstack)
+	local meta = itemstack:get_meta()
+	local metadata = meta:get("")
+	if metadata then
+		meta:set_string("can_level", metadata)
+		meta:set_string("", "")
+		return meta
+	end
+	return meta
+end
--- a/technic/legacy.lua
+++ b/technic/legacy.lua
@@ -47,7 +47,7 @@ technic.legacy_meta_keys = {
 }
 
 -- Converts legacy itemstack metadata string to itemstack meta and returns the ItemStackMetaRef
-function technic.get_stack_meta_compat(itemstack)
+function technic.get_stack_meta(itemstack)
 	local meta = itemstack:get_meta()
 	local legacy_string = meta:get("") -- Get deprecated metadata
 	if legacy_string then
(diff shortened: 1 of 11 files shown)


technic/legacy.lua
```diff
@@ -39,3 +39,45 @@ for i = 0, 64 do
...
+
+-- Meta keys that have changed
+technic.legacy_meta_keys = {
+	["charge"] = "technic:charge",
+}
+
+-- Converts legacy itemstack metadata to itemstack meta and returns the ItemStackMetaRef
+function technic.get_stack_meta_compat(itemstack)
+	local meta = itemstack:get_meta()
+	local metadata = meta:get("")
```
SmallJoker, line 52: Perhaps `legacy_string` would be a better name instead of confusing meta/metadata/metadata_table?

technic/legacy.lua
```diff
@@ -39,3 +39,45 @@ for i = 0, 64 do
...
+		local metadata_table = minetest.deserialize(metadata)
+		if metadata_table then
+			local table = meta:to_table()
+			for k, v in pairs(metadata_table) do
+				local legacy_key = technic.legacy_meta_keys[k]
+				if legacy_key then
+					table.fields[legacy_key] = v
+				else
+					table.fields[k] = v
+				end
```
SmallJoker, line 63:
```suggestion
				table.fields[technic.legacy_meta_keys[k] or k] = v
```

technic/legacy.lua
```diff
@@ -39,3 +39,45 @@ for i = 0, 64 do
 	minetest.register_alias("technic:lv_cable"..i, "technic:lv_cable")
 end
 
+-- Item meta
+
+-- Meta keys that have changed
+technic.legacy_meta_keys = {
+	["charge"] = "technic:charge",
```
SmallJoker, line 46: Why is this translation table needed? Why is the `technic:` prefix not needed for `can_level` below?

technic/legacy.lua
```diff
@@ -39,3 +39,45 @@ for i = 0, 64 do
...
+}
+
+-- Converts legacy itemstack metadata to itemstack meta and returns the ItemStackMetaRef
+function technic.get_stack_meta_compat(itemstack)
```
SmallJoker, line 50: Please rename to `technic.get_stack_meta` that overwrites the trivial non-compat version when `legacy.lua` is loaded. `legacy.lua` should be an optional script for technic.

SmallJoker (commented):
Looks good aside from that. Yet not tested.

technic/legacy.lua
cx384: > Why is this translation table needed?

It is needed to prevent incompatibilities.
e.g. If someone registers an item that can be charged with technic and additional uses a `charge` value for something else.
(It's a table and not just an if statement, since in the future other things may be added, but I can remove the table if you think it is unnecessary.)

> Why is the `technic:` prefix not needed for `can_level` below?

Since `can_level` (unlike power storage) only gets used for cans, and a can (currently) is only a can and can do nothing else. So an arbitrary item can not suddenly get the properties of cans.
(A can can only exist if it has been registered with `technic.register_can(def)` and this directly registers an item.)

Of course the `technic:` prefix could be added to all keys `charge`, `can_level`, `mode`, `target`, `look_depth`, `look_radius`, ...  but I didn't want change them without a good reason.

If you have a different opinion, please tell me how the item meta keys should look like.

technic/legacy.lua
cx384: Done.

technic/legacy.lua
cx384: Changed.

technic/legacy.lua
cx384: Fixed.

technic/legacy.lua
SmallJoker: I see. `charge` is broadly used, thus it makes sense to add a prefix to avoid conflicts. Fine by me as-is.

SmallJoker (approved):
:ballot_box_with_check: Battery box works. Charge gets migrated
:ballot_box_with_check: Can level works and gets migrated
:ballot_box_with_check: Chain saw works and gets migrated
Flashlight works but seems to be bugged (discharges too quickly), not a problem introduced by this PR.
:ballot_box_with_check: Prospector works
:ballot_box_with_check: Sonic screwdriver produces a way too loud sound. It's somewhat weird to use but works.
:ballot_box_with_check: Vacuum cleaner works

I'll merge these changes in a week unless there are objections. This is a longer delay than what I usually do because it affects existing maps.

PR #634, opened 2024-03-09
Add Everness sandstone compressor recipes

gabriel1379:
Hi,

As previously, I wanted to use more Everness stuff with Technic and would like to contribute this too. This time, it's about sand to sandstone and vice versa, using the compressor and the grinder.

I also did a minor refactoring to prevent unnecessary file bloating and preserve overviewability.

Lastly, I needed to add everness as an optional dependency after all, as without it, the default everness mineral_sandstone to mineral_sand recipe could not be cleared. (This should be OK though, as Everness is a more fundamental, world builder mod, while Technic is more for later game, so it shouldn't need to depend on Technic.)

### DIFF
--- a/technic/machines/register/compressor_recipes.lua
+++ b/technic/machines/register/compressor_recipes.lua
@@ -21,6 +21,17 @@ local recipes = {
 	{"technic:uranium35_ingot 5",  "technic:uranium_fuel"},
 }
 
+if minetest.get_modpath("everness") then
+	table.insert(recipes, {"everness:coral_deep_ocean_sand 2",			"everness:coral_deep_ocean_sandstone_block"})
+	table.insert(recipes, {"everness:coral_sand 2",						"everness:coral_sandstone"})
+	table.insert(recipes, {"everness:coral_white_sand 2",				"everness:coral_white_sandstone"})
+	table.insert(recipes, {"everness:crystal_forest_deep_ocean_sand 2", "everness:crystal_forest_deep_ocean_sandstone_block"})
+	table.insert(recipes, {"everness:crystal_sand 2",					"everness:crystal_sandstone"})
+	table.insert(recipes, {"everness:cursed_lands_deep_ocean_sand 2",	"everness:cursed_lands_deep_ocean_sandstone_block"})
+	table.insert(recipes, {"everness:cursed_sand 2",					"everness:cursed_sandstone_block"})
+	table.insert(recipes, {"everness:mineral_sand 2",					"everness:mineral_sandstone"})
+end
+
 -- defuse the default sandstone recipe, since we have the compressor to take over in a more realistic manner
 minetest.clear_craft({
 	recipe = {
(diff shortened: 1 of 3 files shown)


technic/machines/register/compressor_recipes.lua
```diff
@@ -8,6 +8,15 @@ function technic.register_compressor_recipe(data)
 	technic.register_recipe("compressing", data)
 end
 
+function technic.clear_sand_to_sandstone_craft(sand_name)
```
SmallJoker, line 11: Please make this function local. Any new technic API should be documented in doc/ and this one is not special enough to be part of the global technic namespace.

Suggestion: rename the function to what it does internally, such as `clear_shaped_2x2_craft`.

technic/machines/register/compressor_recipes.lua
```diff
@@ -21,25 +30,50 @@ local recipes = {
...
+    local everness_sand_to_sandstone_recipes = {
+        {"everness:coral_deep_ocean_sand 2",          "everness:coral_deep_ocean_sandstone_block"},
+        {"everness:coral_sand 2",                     "everness:coral_sandstone"},
+        {"everness:coral_white_sand 2",               "everness:coral_white_sandstone"},
+        {"everness:crystal_forest_deep_ocean_sand 2", "everness:crystal_forest_deep_ocean_sandstone_block"},
+        {"everness:crystal_sand 2",                   "everness:crystal_sandstone"},
+        {"everness:cursed_lands_deep_ocean_sand 2",   "everness:cursed_lands_deep_ocean_sandstone_block"},
+        {"everness:cursed_sand 2",                    "everness:cursed_sandstone_block"},
+        {"everness:mineral_sand 2",                   "everness:mineral_sandstone"},
+    }
```
SmallJoker, line 43: Please use tabs to indent. Adapt to existing code.

technic/machines/register/compressor_recipes.lua
gabriel1379: Removed and inlined as suggested.

However, for the future, so I'll know, as I'm just getting to know Lua: how exactly do I make a function local? Just write it in front of it, like `local function technic.clear_shaped_2x2_craft` or just leave away the `technic` namespace, like `function clear_shaped_2x2_craft`?

technic/machines/register/compressor_recipes.lua
```diff
@@ -8,6 +8,15 @@ function technic.register_compressor_recipe(data)
 	technic.register_recipe("compressing", data)
 end
 
+function technic.clear_sand_to_sandstone_craft(sand_name)
+	minetest.clear_craft({
```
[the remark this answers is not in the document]
gabriel1379: Good point, added, thx. 👍

technic/machines/register/compressor_recipes.lua
gabriel1379: Seeing how I had issues with tabs (looked differently here than in my editor locally and had to edit it afterwards), I thought spaces would be safer in this regard. But OK, changed back to tabs.

technic/machines/register/compressor_recipes.lua
SmallJoker: `local function mytable.myfunction() end` is not valid. `mytable` is either `local` or global (default).
`function myfunction() end` is a function in the global namespace `_G`. It might overwrite or be overwritten by accident by another mod.
Local variables and functions ( `local function myfunction() end`) can only be seen and used within the current file. This is what I'd recommend for niche utility functions that are unlikely to be of much use outside the file.

Global variables can be shadowed by locals. `luacheck` can partially check against that to avoid side-effects.

technic/machines/register/compressor_recipes.lua
SmallJoker: FYI: most mods that I've seen so far seem to adhere to the Minetest Engine style guideline: https://dev.minetest.net/Lua_code_style_guidelines

Spaces are indeed safer when it comes to aligning *after the first printable character on the line*. The customizable tab size otherwise messes it up.

SmallJoker:
Looks good; the sand recipes are (still) removed as intended.

Will merge in a few days unless anyone's got objections.

technic/machines/register/compressor_recipes.lua
gabriel1379: Ah, I see. Thanks for the explanation. 👍

PR #637, opened 2024-04-07
Feature/add grinder from everness desert stone

gabriel1379:
Hi,

Another little Everness-addition. Everness has its own desert biomes, with their own types of desert stone. Since desert stone is needed to craft the LV grinder, I added the two types of Everness desert stones too as alternative crafting recipes for the LV grinder if the Everness mod is found to be present, so those can be used too (otherwise, if one is using the Everness mod, extremely long trips might be needed to find some `default:desert_stone`.

### DIFF
--- a/technic/machines/LV/grinder.lua
+++ b/technic/machines/LV/grinder.lua
@@ -9,5 +9,25 @@ minetest.register_craft({
 	}
 })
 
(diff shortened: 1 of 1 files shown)


SmallJoker:
LGTM. Will merge in a few days unless there are objections.

PR #638, opened 2024-04-07
Add grinding recipe for nether lump + refactor grinder recipes

gabriel1379:
Hi,

I'm also using the nether mod and wanted to be able to grind nether lumps, so I added a recipe for it.

Also, since I was already at the file and saw that it was getting bigger and harder to overview, refactored it a bit to make it more compact and easier to handle.

(Also, the .png image has been crushed.)


Desour:
Please note that there is more than one nether mod. Using minetest.get_modpath alone is not enough to check for the right one.

gabriel1379:
> Please note that there is more than one nether mod. Using minetest.get_modpath alone is not enough to check for the right one.

Thanks for the hint. Additional check added (also to the other PR). Please check and let me know if that's OK or if anything else is needed.

Desour:
>Please check and let me know if that's OK or if anything else is needed.

To be clear, I'm not a maintainer of this mod, I just wanted to make sure it doesn't get incompatible with some mods. :)

gabriel1379:
> > Please check and let me know if that's OK or if anything else is needed.
> 
> To be clear, I'm not a maintainer of this mod, I just wanted to make sure it doesn't get incompatible with some mods. :)

Ah, ok. Still, thanks for the hint. :)

PR #639, opened 2024-04-08
Feature/add compressor recipes for nether brick and lump

gabriel1379:
Hi,

Another addition with nether, similar to the previous one, just this time for compressor.

Clears nether's default recipes for compressed nether brick and nether lump, using the compressor instead for these tasks.

+ did the same type of clearing up for better overview as with the grinder recipes in the other MR.

### DIFF
--- a/technic/machines/register/compressor_recipes.lua
+++ b/technic/machines/register/compressor_recipes.lua
@@ -11,7 +11,7 @@ end
 -- Function for cases where multiple mods exist with the same name,
 -- to check if we have the right one.
 local function is_right_mod(dependency)
-	local is_right = true
+	is_right = true
 
 	if dependency == "nether" then
 		is_right = minetest.registered_nodes["nether:brick_compressed"] ~= nil
--- a/technic/machines/register/compressor_recipes.lua
+++ b/technic/machines/register/compressor_recipes.lua
@@ -38,7 +38,18 @@ if minetest.get_modpath("everness") then
 	end
 end
 
--- defuse the default sandstone recipe, since we have the compressor to take over in a more realistic manner
+if minetest.get_modpath("nether") then
+	local nether_brick_and_lump_recipes = {
+		{"nether:brick 9",				"nether:brick_compressed"},
+		{"nether:brick_compressed 9",	"nether:nether_lump"},
+	}
+
+	for _, data in ipairs(nether_brick_and_lump_recipes) do
+		table.insert(recipes, {data[1], data[2]})
+	end
+end
+
+-- Defuse the default sandstone recipes, since we have the compressor to take over in a more realistic manner.
 local crafts_to_clear = {
 	"default:desert_sand",
 	"default:sand",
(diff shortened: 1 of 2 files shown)


technic/machines/register/compressor_recipes.lua
```diff
@@ -55,23 +38,92 @@ if minetest.get_modpath("everness") then
...
+		for _, craft_entry in ipairs(crafts) do
+			table.insert(crafts_to_clear, craft_entry)
+		end
 	end
 end
 
-for _, sand_name in ipairs(crafts_to_clear) do
+-- Clear recipes
+for _, craft_name in ipairs(crafts_to_clear) do
+	is_regular = string.sub(craft_name, 1, 7) ~= "nether:"
```
SmallJoker, line 60: Please do not pollute the global namespace. Use `local is_regular`. Same for other occurrences such as `shaped_recipe`.

technic/machines/register/compressor_recipes.lua
```diff
@@ -8,45 +8,28 @@ function technic.register_compressor_recipe(data)
...
-		{"everness:coral_white_sand 2",               "everness:coral_white_sandstone"},
-		{"everness:crystal_forest_deep_ocean_sand 2", "everness:crystal_forest_deep_ocean_sandstone_block"},
-		{"everness:crystal_sand 2",                   "everness:crystal_sandstone"},
-		{"everness:cursed_lands_deep_ocean_sand 2",   "everness:cursed_lands_deep_ocean_sandstone_block"},
-		{"everness:cursed_sand 2",                    "everness:cursed_sandstone_block"},
-		{"everness:mineral_sand 2",                   "everness:mineral_sandstone"},
-	}
+-- Function for cases where multiple mods exist with the same name,
+-- to check if we have the right one.
+local function is_right_mod(dependency)
```
SmallJoker, line 13: Why exactly do we need this function? Uncraftable recipes do not pose any significant performance impact - at least not any more (fixed in recent Minetest versions). You could register the recipes as soon as `minetest.get_modpath` returns a non-nil value. That would also get rid of the `nether` dependency.

gabriel1379: User "Desour" on my other pull request ( https://github.com/minetest-mods/technic/pull/638 ) recommended an additional check for making sure that the right nether mod is being used. But OK, removed it now.

However, the dependency is still needed as recipes from nether (the compressing ones) still need to be cleared too.

technic/machines/register/compressor_recipes.lua
gabriel1379: Right - I keep forgetting that this needs to be noted separately in Lua. Thanks for the reminder, made local.

technic/machines/register/compressor_recipes.lua
gabriel1379: Removed it from the other PR too just now.

SmallJoker (approved):
Looks good. Sandstone, desert stone and silver sandstone are still craftable.

Will merge in a few days unless there are objections.

PR #644, opened 2024-06-16
Add compressor recipes for nether racks

gabriel1379:
Hi,

Noticed in [the discussion over at nether](https://github.com/minetest-mods/nether/issues/77) that these recipes were missing, so added them now.


SmallJoker (approved):
LGTM. Will merge in a few days unless there are objections.

PR #645, opened 2024-07-03
Expose get/set charge functions

OgelGames:
Exposes the battery box's get/set charge functions, for easier compatibility with other mods and other versions of technic.


SmallJoker:
I know you did not write this function but documenting it would be very much appreciated. Otherwise it'll end up being a function where we don't know whether it's exposed intentionally or just an oversight.

SmallJoker:
Thank you very much. I will merge this PR in a few days unless there are objections.

PR #651, opened 2024-09-28
Radiation API

### LINKED ISSUE #605: Ability to declare nodes as radiation protection and corium-resistant
The list of radiation shielding nodes is hardcoded and cannot be changed by external mods. I suggest either adding a group or API method for registering custom radiation protection nodes

For corium, I suggest adding a group that makes nodes in that group unaffected by corium

DustyDave961:
Radiation resistance API made with the help of Microsoft Copilot to fix #605.


DustyBagel (commented):
Seems fine. This should probably be reviewed by someone who understands how this works better though.

technic/radiation.lua
```diff
@@ -28,9 +28,26 @@ or complex internal structure should show no radiation resistance.
...
+    rad_resistance_node[node_name] = resistance
+    cache_radiation_resistance[node_name] = nil -- Invalidate cache
+end
+
+-- Function to register multiple node resistances at once
+function technic.register_multiple_resistances(resistances)
+    for node_name, resistance in pairs(resistances) do
+        technic.register_rad_resistance(node_name, resistance)
+    end
+end
```
SmallJoker, line 46: Functions should be kept as generic as possible. This one can be very easily be implemented by modders. It does not bring any noticeable benefit.

technic/radiation.lua
```diff
@@ -28,9 +28,26 @@ or complex internal structure should show no radiation resistance.
 Fractional resistance values are permitted.
 --]]
 
+local rad_resistance_node = {}
+local rad_resistance_group = {}
+local cache_radiation_resistance = {}
+
+-- Function to register node-specific resistance
+function technic.register_rad_resistance(node_name, resistance)
```
SmallJoker, line 36: Please document this function in `doc/api.md`.

technic/radiation.lua
```diff
@@ -165,38 +181,62 @@ local rad_resistance_node = {
...
+    end
 end
 
+technic.register_group_resistance("concrete", 16)
+technic.register_group_resistance("tree", 3.4)
+technic.register_group_resistance("uranium_block", 500)
+technic.register_group_resistance("wood", 1.7)
+
+-- Function to calculate radiation resistance
+function node_radiation_resistance(node_name)
```
SmallJoker, line 205: This function should either be local, or in the `technic.` table and documented.

technic/radiation.lua
```diff
@@ -165,38 +181,62 @@ local rad_resistance_node = {
...
+        return 0
+    end
+
+    -- Check for rad_resistance group in node definition
+    resistance = 0
+    for g, v in pairs(def.groups) do
+        if g == "rad_resistance" then
+            resistance = resistance + v
+        end
+    end
```
SmallJoker, line 222: `technic.register_group_resistance` should be used instead of this special case.

technic_worldgen/nodes.lua
```diff
@@ -106,7 +106,7 @@ minetest.register_node(":technic:lead_block", {
 	description = S("Lead Block"),
 	tiles = { "technic_lead_block.png" },
 	is_ground_content = true,
-	groups = {cracky=1, level=2},
+	groups = {cracky=1, level=2, rad_resistance=80},
```
SmallJoker, line 109: `technic.register_rad_resistance` could be used here. Or register a group.

technic/radiation.lua
DustyDave961: I actually didn't change the name of that one since it's used to calculate radiation damage later on, but you do have a point.

technic/radiation.lua
DustyDave961: How would that even work? The whole point of the rad_resistance group is for individual nodes to have radiation resistance built into their node definitions. The group registration is for groups of nodes that share the same resistance value.

technic_worldgen/nodes.lua
DustyDave961: Adding it to the node definitions seemed simpler for new nodes. I figured people could do it whatever way they wanted. Should I have stuck with just the registration functions?

technic/radiation.lua
DustyDave961: I suppose that's fair.

technic/radiation.lua
DustyDave961: I put it under the Helper functions. Will that work, or should it have its own section?

technic/radiation.lua
SmallJoker: I was thinking of using the formula `resistance = rad_resistance_group[g] * v` where g is the group name and v the group value. The group value would multiply the base resistance.

That idea is however flawed. Group values can be used for other purposes, such as tool capabilities, thus have no relation to the radiance resistance.

Consider my original comment as invalid.

technic/radiation.lua
SmallJoker: Fine by me.

technic_worldgen/nodes.lua
SmallJoker: This comment comes from the same underlying [flawed] thought as https://github.com/minetest-mods/technic/pull/651#discussion_r1801732331 .

This group is OK as-is.

DustyDave961: I suppose one reason I did it that way is because I haven't learned how to override groups without overwriting the original groups. Registering group resistances with the rad_resistance group sounds even more complicated.

technic/radiation.lua
```diff
@@ -147,57 +158,49 @@ local rad_resistance_node = {
...
+				node_def.groups.rad_resistance = resistance
+			end
+		end
 	end
+end
+
+technic.register_group_resistance("concrete", 16)
+technic.register_group_resistance("tree", 3.4)
+technic.register_group_resistance("uranium_block", 500)
+technic.register_group_resistance("wood", 1.7)
```
SmallJoker, line 186: Unfortunately this only works on nodes that are already registered when `technic` is loaded.
A few ideas to fix that:
1. re-add the previous `cache_radiation_resistance` to lazily calculate the node value when needed
2. calculate all values either in `core.register_on_mods_loaded` or in a `core.after(0, function() ??? end)`  callback (first server step)

technic/radiation.lua
```diff
@@ -28,9 +28,20 @@ or complex internal structure should show no radiation resistance.
...
+-- Function to register node-specific resistance
+function technic.register_rad_resistance(node_name, resistance)
+	local node = minetest.registered_nodes[node_name]
+	if node then
+		if not node.groups then
+			node.groups = {}
+		end
+        node.groups.rad_resistance = resistance
+	end
+end
```
SmallJoker, line 40: FYI: (no action needed in this case)
For this group it does not matter, but you'd generally want to use `core.override_item(node_name, { groups = newgroups })` so that the Minetest engine is aware of the groups as well. It is important for the client-side prediction of armor and tool digging groups (i.e. to determine how long it takes to dig a node with the selected tool).

technic_worldgen/nodes.lua
SmallJoker: It always depends on how much code you're saving vs the amount of care the code needs for maintenance and documentation. In this case I think it is OK to have a shorthand.

technic/radiation.lua
DustyDave961: I found such functions in the Minetest lua api, but I'm unsure how to use the `minetest.register_on_mods_loaded(function())` function. Do I just replace "function" `with node_radiation_resistance`, or is there more to it?

technic/radiation.lua
DustyDave961: Does `minetest.override_item` remove groups that aren't included in the new groups list? I didn't use it because I was concerned about that.

technic/radiation.lua
SmallJoker: @DustyDave961 [Changelog `0.4.16 → 5.0.0`](https://dev.minetest.net/Changelog) mentions this callback. Minetest 5.0.0 is already the minimal requirement for technic, thus this is not of a concern.

If functions are not mentioned in the changelog, you might as well have a look at the git blame of the `lua_api.md` line, which reveals when it was added.

technic/radiation.lua
SmallJoker: `core.override_item` overrides the fields you're giving it, so here's how you can preserve them:

```Lua
local groups  = core.registered_nodes[node_name].groups or {}
groups.mygroupname = 123
core.override_item(node_name, groups)
```

DustyDave961: Thank you, that makes sense to me.

PR #654, opened 2024-11-16
power monitor instability fix

### LINKED ISSUE #640: Power Monitors Unstable State
The power monitor when connected to a valid power network swaps between the connected and disconnected states.

SomeName42:
Fixed power monitor instability where it was switching between having and not having network infotext. Resolves #640. The cause of the issue was that the power monitor is LV MV and HV machine at the same time, and each tier has a countdown. If any of these countdowns were 0 it treated the machine as having no network. This was changed so that it only treats a machine as having no network if it is timed out in all tiers that it is a part of.

### DIFF
--- a/technic/machines/switching_station.lua
+++ b/technic/machines/switching_station.lua
@@ -468,20 +468,27 @@ minetest.register_abm({
 	interval   = 1,
 	chance     = 1,
 	action = function(pos, node, active_object_count, active_object_count_wider)
+		local has_network = false
+		local technic_machine = false
 		for tier, machines in pairs(technic.machines) do
-			if machines[node.name] and switching_station_timeout_count(pos, tier) then
-				local nodedef = minetest.registered_nodes[node.name]
-				if nodedef then
-					local meta = minetest.get_meta(pos)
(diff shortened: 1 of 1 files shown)


technic/machines/switching_station.lua
```diff
@@ -468,22 +468,31 @@ minetest.register_abm({
...
+				technic_machine = true
+				if not switching_station_timeout_count(pos, tier) then
+					has_network = true
 				end
 			end
 		end
+
+		if technic_machine and not has_network then
+			local nodedef = minetest.registered_nodes[node.name]
+			if nodedef then
```
SmallJoker, line 484: This ABM depends on the node group, which means that the node must be registered. Hence such check should always evaluate to `true`. If not, then it's probably an engine bug.

SomeName42: Do you mean the technic_machine variable?

SmallJoker: No, I mean the line that I commented on. `if nodedef then` should (in this case) be equal to `if true then`.

PR #655, opened 2024-11-16
power multiplication fix

### LINKED ISSUE #277: Power Doubling Bug
It's hard for me to clearly explain in words on how to reproduce it so I will just post a whole bunch of pictures.

It is possible to abuse this to generate six times as much energy. For example, using the nuclear generator, that means 600000, although the alarm would be ringing and it would melt down.

screenshot_2016-04-22_21-02-34
the first energy update seem to have some kind of race condition or something...
screenshot_2016-04-22_21-02-41
screenshot_2016-04-22_21-02-52
screenshot_2016-04-22_21-03-00
screenshot_2016-04-22_21-03-08
screenshot_2016-04-22_21-03-17
last image in another angle, to clearly show how to build it.
screenshot_2016-04-22_21-03-49

____
Note: bug is **NOT** fix by https://github.com/minetest-mods/technic/commit/33455328bd4fbfacd26e3f02b77f2b54bea982d9
See https://github.com/minetest-mods/technic/issues/277#issuecomment-293822014

SomeName42:
Fixed power multiplication bug. Resolves #277. The cause of the bug was that machines can be a part of multiple networks and generate power for each of the networks. This was fixed by checking if machine is already part of an other network and not making it a part of the current one. When network times out the network meta string value is cleared and the clear_networks function is used to try to connect the machine to a different network.

### DIFF
--- a/technic/machines/register/cables.lua
+++ b/technic/machines/register/cables.lua
@@ -43,6 +43,8 @@ local function check_connections(pos)
 	return connections
 end
 
+local clear_networks
+
 local function clear_network(net)
 	for _,v in pairs(technic.networks[net].all_nodes) do
 		local pos1 = minetest.hash_node_position(v)
@@ -62,7 +64,7 @@ local function clear_network(net)
 	end
 end
 
-local function clear_networks(pos)
+clear_networks = function(pos)
 	local node = minetest.get_node(pos)
 	local meta = minetest.get_meta(pos)
 	local placed = node.name ~= "air"
--- a/technic/machines/register/cables.lua
+++ b/technic/machines/register/cables.lua
@@ -45,13 +45,17 @@ end
 
 local clear_networks
 
-local function clear_network(net)
-	for _,v in pairs(technic.networks[net].all_nodes) do
+local function clear_network(network_id)
+	if not network_id then
+		return
+	end
+
+	for _,v in pairs(technic.networks[network_id].all_nodes) do
 		local pos1 = minetest.hash_node_position(v)
 		technic.cables[pos1] = nil
 	end
-	local network_bckup = technic.networks[net]
-	technic.networks[net] = nil
+	local network_bckup = technic.networks[network_id]
+	technic.networks[network_id] = nil
 
 	for _,n_pos in pairs(network_bckup.PR_nodes) do
 		clear_networks(n_pos)
@@ -86,7 +90,6 @@ clear_networks = function(pos)
 
 			-- Actually add it to the (cached) network
 			-- !! IMPORTANT: ../switching_station.lua -> check_node_subp() must be kept in sync
-			pos.visited = 1
 			if technic.is_tier_cable(node.name, tier) then
 				technic.cables[minetest.hash_node_position(pos)] = network_id
 				table.insert(network.all_nodes,pos)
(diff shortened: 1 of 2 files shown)


technic/machines/switching_station.lua
```diff
@@ -100,9 +100,11 @@ end
 -- Add a wire node to the LV/MV/HV network
 -- Returns: indicator whether the cable is new in the network
 local hash_node_position = minetest.hash_node_position
-local function add_network_node(nodes, pos, network_id)
+local function add_network_node(nodes, pos, network_id, cable)
```
SmallJoker, line 103: 1. Optimization idea: `network_id` could be set to `nil` for non-cables.
2. Line 100 needs an update. It adds all kinds of nodes; not just cables.

technic/machines/register/cables.lua
```diff
@@ -43,84 +43,108 @@ local function check_connections(pos)
...
-				technic.networks[net] = nil
 			end
 		end
+		return
+	end
+
+	local net = technic.cables[minetest.hash_node_position(pos)]
+	if net and technic.networks[net] then
+		clear_network(net)
+	end
```
SmallJoker, line 140: The name `net` does not clarify the variable contents. Furthermore, `t[nil]` cannot be assigned, thus the condition can be abbreviated:
```suggestion
	local network_id = technic.cables[minetest.hash_node_position(pos)]
	if technic.networks[network_id] then
		clear_network(network_id)
	end
```
.. or maybe have a guard in `clear_network` so that you can pass it directly without checking here.

technic/machines/register/cables.lua
```diff
@@ -43,84 +43,108 @@ local function check_connections(pos)
...
+
+	for _,connected_pos in pairs(positions) do
+		local net = technic.cables[minetest.hash_node_position(connected_pos)]
+		if net and technic.networks[net] then
```
SmallJoker, line 144: Same comment as a few lines further up.

technic/machines/switching_station.lua
```diff
@@ -136,21 +138,31 @@ local check_node_subp = function(network, pos, machines, sw_pos, from_below, net
...
+	local network_str = string.format("%.20g", network_id)
+	local network_key = network.tier.."_network"
+	local m_network_str = meta:get_string(network_key)
+
+	if m_network_str == "" then
+		meta:set_string(network_key, network_str)
+	else
+		if m_network_str ~= network_str then
+			return
+		end
```
SmallJoker, line 150: When is `return` executed? Wouldn't this indicate a network logic error? Hence: would a warning be appropriate to find such mismatches?

technic/machines/register/cables.lua
```diff
@@ -43,84 +43,108 @@ local function check_connections(pos)
 	return connections
 end
 
-local function clear_networks(pos)
+local clear_networks
+
+local function clear_network(net)
```
SmallJoker, line 48: `net` -> `network_id` to clarify the variable contents.

technic/machines/register/cables.lua
```diff
@@ -43,84 +43,108 @@ local function check_connections(pos)
...
+			if not network_id then
+				-- We're evidently not on a network, nothing to add ourselves to
+				return
+			end
+			local network = technic.networks[network_id]
+			local tier = network.tier
+
+			-- Actually add it to the (cached) network
+			-- !! IMPORTANT: ../switching_station.lua -> check_node_subp() must be kept in sync
+			pos.visited = 1
```
SmallJoker, line 89: This field/assignment could be removed. It is unused and not documented anywhere.

SmallJoker (commented):
Works well so far. Connected and disconnected a few cables. The machines are updated correctly.

technic/machines/switching_station.lua
SomeName42: return is executed when the machine is on a different network already therefore it is ignored

SmallJoker (approved):
Thank you for the update.
Works. Will merge in a few days unless there are objections.

PR #661, opened 2025-02-25
Fix shading on CNC shape models

Montandalar:
Luanti 5.11 showed a lot of flaws in the shading of the CNC models, some of which were previously obscured. Fix the models so that smooth shading is applied where appropriate and sharp/flat shading where appropriate. The modified files have also been run through ExeVirus' CompressObj.

Full notes on changes in the spreadsheet. See the before and after image.

[Mainline-Models-Fixup.ods](https://github.com/user-attachments/files/18958449/Mainline-Models-Fixup.ods)
mainline_shapes_before_after


SmallJoker (approved):
Thank you for the corrections. Looks good!
Will merge in a few days unless there are objections.

PR #662, opened 2025-05-02
support every paramtype2

The4codeblocks:


technic/tools/sonic_screwdriver.lua
```diff
@@ -50,18 +60,54 @@ local function screwdriver_handler(itemstack, user, pointed_thing, mode)
...
+
+		if mode == ROTATE_FACE then
+			rotation = rotation + 1
+		elseif mode == ROTATE_AXIS then
+			rotation = rotation + 4
+		end
+		rotation = rotation % 24
+
+		new_param2 = aux * 32 + rotation
 	end
```
SmallJoker, line 108: Idea: You could move this code up such that a final `else` case could be used for `return`. It would effectively deduplicate the lines L38-47.

Admittedly this code does not look great but I don't see much that could be done to improve it.

The4codeblocks: should be done

technic/tools/sonic_screwdriver.lua
```diff
@@ -31,16 +31,80 @@ local function screwdriver_handler(itemstack, user, pointed_thing, mode)
...
+		end
+		new_param2 = new_param2 % 240
+	elseif (paramtype2 == "colordegrotate") then
+		local aux = floor(param2 / dir_component)
+		local rotation = param2 % dir_component
+
+		if mode == ROTATE_FACE then
+			rotation = rotation + 1
+		elseif mode == ROTATE_AXIS then
+			rotation = rotation + 4
```
SmallJoker, line 99: When I use `paramtype2 = "degrotate",` in the nodedef of `"default:aspen_sapling"` (nodes.lua), the rotation oddly jumps in this case. There is no such anomaly with `paramtype2 = "degrotate",`.
Is it because you're using modulo 32 in L94, but modulo 24 in L101?

The4codeblocks: describe the jump
(%32 is bitwise extraction, %24 is rotation modulo)

SmallJoker: Sorry. I meant that `colordegorate` jumps oddly whereas `degrotate` works well.
It turns out this code is not flawed but my perception. It simply turned too far to notice at a stationary viewing position. When following the node around in its rotation, there is no such "glitch" or "jump".
Resolved.

technic/tools/sonic_screwdriver.lua
```diff
@@ -49,19 +113,7 @@ local function screwdriver_handler(itemstack, user, pointed_thing, mode)
...
 	minetest.sound_play("technic_sonic_screwdriver", {pos = pos, gain = 0.3, max_hear_distance = 10})
 
-	-- Set param2
-	local rotationPart = node.param2 % 32 -- get first 4 bits
-	local preservePart = node.param2 - rotationPart
-
-	local axisdir = math.floor(rotationPart / 4)
-	local rotation = rotationPart - axisdir * 4
-	if mode == ROTATE_FACE then
-		rotationPart = axisdir * 4 + nextrange(rotation, 3)
```
SmallJoker, line 59: luacheck is yet unhappy `technic/tools/sonic_screwdriver.lua:11:16: unused function nextrange`
To fix that, you could ignore the warning (I hope it's the correct code) or remove the function.
```Lua
-- luacheck: push ignore 211
local function foobar()
end
-- luacheck: pop
```

The4codeblocks: > remove the function

done: https://github.com/minetest-mods/technic/pull/662/commits/150e738b12be187d50152421eeaede071096c18a

SmallJoker:
Thanks. Looks good. Will merge in a few days unless there are objections.

PR #665, opened 2025-08-25
Start using settingtypes.txt

DustyBagel:
This currently isn't finished but should be soon. I need help figuring out how to correctly generate the technic.config.

### DIFF
--- a/settingtypes.txt
+++ b/settingtypes.txt
@@ -7,3 +7,30 @@
 #
 #    Disabling this feature will sacrifice safety for convenience.
 technic_safe_chainsaw (Chainsaw safety feature) bool true
+
+# Enables the mining drills.
+enable_mining_drill (Enable Mining Drill) bool true
+
+# Enables Wind Mills.
+enable_wind_mill (Enable Wind Mill) bool false
+
+# Enables the Flashlight.
+enable_ (Enable Flashlight) bool false
+
+# Enables Frames.
+enable_frames (Enable Frames) bool false
+
+# Enables Corium Griefing.
+enable_corium_griefing (Enable corium_griefing) bool true
+
+# Enables Radiation Protection.
+enable_radiation_protection (Enable Radiation Protection) bool true
+
+# Enables Entity Radiation Damage.
+enable_entity_radiation_damage (Enable Entity Radiation Damage) bool true
+
+# Enables Long Term Radiation Damage.
+enable_longterm_radiation_damage (Enable Long Term Radiation Damage) bool true
+
+# Enables Nuclear Reactor Digiline Selfdestruct.
+enable_nuclear_reactor_digiline_selfdestruct (Enable Nuclear Reactor Digiline Selfdestruct) bool false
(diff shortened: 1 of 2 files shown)


settingtypes.txt
```diff
@@ -7,3 +7,30 @@
 #
 #    Disabling this feature will sacrifice safety for convenience.
 technic_safe_chainsaw (Chainsaw safety feature) bool true
+
+# Enables the mining drills.
```
SmallJoker, line 11: `Enable Mining Drill` is already shown in the settings. Players expect more information if a help icon is shown. If there is no additional information to show, then please remove these description lines.

-->
```
enable_mining_drill (Enable Mining Drill) bool true

enable_wind_mill (Enable Wind Mill) bool false
```

settingtypes.txt
```diff
@@ -7,3 +7,30 @@
...
+enable_mining_drill (Enable Mining Drill) bool true
+
+# Enables Wind Mills.
+enable_wind_mill (Enable Wind Mill) bool false
```
SmallJoker, line 15: Please use the `technic_` prefix on all settings such that that their origin is clear. The minetest.conf file is already somewhat chaotic, thus let's not make it much worse.

settingtypes.txt
DustyBagel: I know. I just wasn't sure what i should put there. The descriptions should be written by someone who understands technic more. I suppose I could remove them.

settingtypes.txt
DustyBagel: Good idea. Will do.

technic/config.lua
```diff
@@ -13,10 +20,16 @@ local defaults = {
 	enable_entity_radiation_damage = "true",
 	enable_longterm_radiation_damage = "true",
 	enable_nuclear_reactor_digiline_selfdestruct = "false",
+	technic_safe_chainsaw = "true",
 }
 
 for k, v in pairs(defaults) do
 	if conf_table[k] == nil then
-		technic.config:set(k, v)
+		local minetest_val = minetest.settings:get(k)
```
SmallJoker, line 28: You could use `technic.config:set(k, core.settings:get_bool(k, v))` where `v` is the fallback. Requires Minetest 5.0.0 or newer. This is already a requirement, thus usable without further adjustments.

DustyBagel: I didn't put the nil check for compatibility. If we call one of the core.settings functions and the setting is at it's default value then the function will return nil anyway.

SmallJoker: Which part would return nil?

Correction of my previous example:
```Lua
technic.config:set(k_but_without_technic_prefix, tostring(core.settings:get_bool(k, v)))
-- can be either "true" or "false"
```

DustyBagel: I seem to recall that core.settings:get and core.settings:get_bool return nil when the setting hasn't been changed at least once?

SmallJoker: Yes, without the 2nd argument it would indeed return `nil`. However, providing the fallback argument will act as a default value.

PR #671, opened 2026-06-03
quecto, ronto, ronna, quetta

The4codeblocks:


SmallJoker:
Where are such high numbers used? This seems somewhat out-of-range what is applicable to `technic`.

The4codeblocks:
> Where are such high numbers used? This seems somewhat out-of-range what is applicable to `technic`.

they're used as often as zetta and yotta, which are part of technic
this pr is for completeness

SmallJoker:
> used as often as zetta and yotta

Hence, not at all. EU is usually below 1 million and I'm not aware of any use of this function other than that.

The4codeblocks:
> > used as often as zetta and yotta
> 
> Hence, not at all.

so the reason zetta and yotta are included is not that they'd be used, it's for completeness

SmallJoker:
Adopted in 2022. https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes
At least that list is finite.

Issue #656, opened 2024-11-18
Unexpected Conversion of Stainless Steel to Lead


Bituvo:
The technic manual says to use stainless steel blocks for one of the layers of a nuclear reactor. So I, not knowing it would all be converted into lead blocks, built it, and was surprised when it happened. The relevant code appears to be [here](https://github.com/minetest-mods/technic/blob/18df2813a0abfe6cf890223254cb2286260b51d5/technic/machines/HV/nuclear_reactor.lua#L206). The reactor should work with either stainless steel or lead blocks if this is the case. I lost a lot of stainless steel this way.

DustyBagel:
I remember experiencing the same thing. It's annoying but a minor bug. I agree that the stainless steel blocks should not turn into lead.

SmallJoker:
Removed the conversion code. A proper implementation is to introduce a node metadata field for freshly placed reactor cores and decide the shell/structure material behaviour based on that.

b6ef3a2

PR #674, opened 2026-06-27
Manual: Use lead OR stainless steel in reactors

TangentFoxy:
Because the mod will convert stainless steel into lead, tell people about the option and warn them about the change. This will mitigate #656 until a decision is made about conversion.


SmallJoker:
Removed the conversion code in b6ef3a2.

TangentFoxy:
Should the manual be changed to indicate there's a difference depending on the version?

SmallJoker:
I think documenting the newest behaviour should be sufficient. (i.e. use Lead Blocks, where Stainless Steel Blocks serve as a fall-back and might be deprecated in the future).

SmallJoker:
Thank you.

PR #678, opened 2026-07-31
Fix circular dependency with technic_worldgen

Darin755:
I was unable to load Technic on Luanti 5.16.1 due to a circular dependency inside Technic. I'm not exactly sure what changed in Luanti but for whatever reason the dependency checker now apparently checks optional_dependencies which creates a dependency loop. This PR just deletes technic from technic_worldgen to solves the issue.


SmallJoker:
Sorry, I do not know how this happened. I did test the `technic_worldgen` changes in a world. Perhaps I did not enable `technic` (?) because that would be the only explanation.

Thank you for the fix.

PR #679, opened 2026-07-31
Break dependency chains

SmallJoker:
Mods that are not required for the functionality of the mod (which includes recipes) are now removed. Trivial mods are kept due to low risk of circular dependencies.

This is a follow-up on #678


SmallJoker:
@Darin755 If you happen to have the chance - would you please be so nice to check whether this works as expected?

extranodes/init.lua
```diff
@@ -2,8 +2,53 @@
 -- namespace: technic
 -- Boilerplate to support localized strings if intllib mod is installed.
 local S = rawget(_G, "intllib") and intllib.Getter() or function(s) return s end
+local stairsplus = core.get_modpath("moreblocks") and stairsplus
```
Desour: This will break if moreblocks isn't loaded first.

Desour (commented):
I always thought of optional dependencies not only as a mean to control load order but also to communicate to users which mods are used.

SmallJoker:
> I always thought of optional dependencies not only as a mean to control load order but also to communicate to users which mods are used.

I agree. Unfortunately there's no way to tell whether the dependency is required to load *before* the current mod, which is the main reason why I am trying to move this information to the README (which is unfortunately not easily seen by players).

extranodes/init.lua
SmallJoker: Indeed. That's why it's still listed as optional dependency.

Desour: Oh, I must've been blind.

But the issue also exists for `digiline_remote` in `technic`, and there the opt dep is missing (unless I'm blind once again).

SmallJoker: Thanks. That's indeed a mistake. Fixes.

